PR #5, opened 2022-01-28
Adding FASTQC Support

### LINKED ISSUE #4: Implement FASTQC feature
As of right now, YEAT does not have a FASTQC feature.

danejo3:
The purpose of this PR is to resolve #4.

### DIFF
--- a/yeat/Snakefile
+++ b/yeat/Snakefile
@@ -13,7 +13,8 @@ import pandas as pd
 
 rule all:
     input:
-        expand("analysis/quast/{sample}/report.html", sample=config["sample"])
+        expand("analysis/quast/{sample}/report.html", sample=config["sample"]),
+        expand("seq/fastqc/{sample}_R{number}_fastqc.html", sample=config["sample"], number=["1", "2"])
 
 
 rule copyinput:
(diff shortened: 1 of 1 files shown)


danejo3:
I haven't been able to figure out how to remove this warning. This warning is also seen in `mystt` as well when you run the tests.

It seems like the deprecation warning is coming from a library that python uses within itself. I am using python 3.9.7.

I found a thread of someone pointing this out and creating a PR for it. (Not sure what rabbit hole I went down into...)

https://github.com/RazerM/ratelimiter/issues/10

image

standage:
I've seen it as well. Deprecation means that a language feature is slated for removal, but hasn't actually been removed yet. A kind of grace period during which packages have time to update the offending behavior. In both cases, the culprit code is in third-party libraries, so there's not much we can do except hope that the libraries are updated before the warnings become errors in the future.

danejo3:
@standage I think I've added everything for `fastqc` report. Let me know if you have any questions, thoughts, or concerns with this PR.

yeat/Snakefile
```diff
@@ -28,6 +29,19 @@ rule copyinput:
...
+    output:
+        out1="seq/fastqc/{sample}_R1_fastqc.html",
+        out2="seq/fastqc/{sample}_R2_fastqc.html"
+    input:
+        read1="seq/input/{sample}_R1.fq.gz",
+        read2="seq/input/{sample}_R2.fq.gz"
+    shell:
+        """
+        fastqc {input.read1} {input.read2} -o seq/fastqc/
+        """
```
standage, line 42: Only potential comment: might want to add `threads: 2` to the rule and `--threads {threads}` to the shell command for parallel execution.

standage (changes requested):
Looks good overall!

standage:
A non-dry-run test of the "Animal" data bundled with YEAT was successful.

danejo3:
Thanks for the comments. Added everything that was suggested.

standage (approved):

Issue #10, opened 2022-03-21
Adding support for MEGAHIT


danejo3:
[MEGAHIT](https://github.com/voutcn/megahit) is an ultra-fast and memory-efficient (meta-)genome assembler.

This assembler would be another option that users can run with their reads for memory efficiency.

In a [review](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5502489/#:~:text=While%20SPAdes%20was%20the%20best,a%20fraction%20of%20computational%20resources.) paper comparing the various meta-genome assemblers out there, the authors wrote,

> Overall, SPAdes, metaSPAdes, IDBA-UD and MEGAHIT displayed the best performances in assembling this metagenome of intermediate size and complexity, as they produced very high N50 values, a high proportion of long contigs and the widest assembly spans. **While SPAdes was the best assembler overall, MEGAHIT was the most memory efficient, as it produced an assembly comparable to the best performing assemblers while using only a fraction of computational resources.**

When deciding on which assembler to use based on the total RAM available, it is suggested that MEGAHIT is the choice assembler if there are < 500 GB of RAM available.

In addition to comparing MEGAHIT to SPADES, MEGAHIT is folds-faster than metaSPADES and SPADES.

Depending on time constraints, MEGAHIT is a great option over SPADES with minimal losses to the overall genome assembly.

PR #12, opened 2022-03-23
Implementing a config file for handling various assembly algorithms

danejo3:
The purpose of this PR is to lay down the foundations of #10 .

In this PR, users will now be required to provide a `config` file that contains information on what assembly algorithms that are to be used.

The reason why having this `config` file would be helpful is to allow users to run multiple assemblers in the future asynchronously. By having multiple assembly algorithms running at the sametime, this will allow the user to work with various assemblers based on time, resource, and accuracy constraints.

This PR does not provide the parallel jobs support and additional assemblers such as `MEGAHIT`. Separate PR will be created for these upcoming features.


danejo3 (commented):
testing

yeat/assembly/config.py
```diff
@@ -0,0 +1,44 @@
...
+            algorithm = d["algorithm"]
+            if algorithm not in ASSEMBLY_ALGORITHMS:
+                message = f"Found unsupported assembly algorithm in config file: [[{algorithm}]]!"
+                raise AssemblyConfigError(message)
+            if algorithm in algorithms:
+                message = f"Found duplicate assembly algorithm in config file: [[{algorithm}]]!"
+                raise AssemblyConfigError(message)
+            algorithms.append(algorithm)
+            assemblers.append(AssemblyConfiguration.from_json(d))
+        return assemblers
```
danejo3, line 44: The purpose of the `parse_json()` is to not only parse the config file but also scan for any invalid or duplicate inputs made by the user. 

For example, 
- an assembly algorithm's name is either misspelled or is not supported by YEAT
- the user added duplicate assembly algorithms in the config file

yeat/cli/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+from pkg_resources import resource_filename
+from snakemake import snakemake
+from yeat.assembly import config
+
+
+SNAKEFILES = {"spades": resource_filename("yeat", "assembly/Snakefile_Spades")}
+
+
+def run(read1, read2, assembler, outdir=".", cores=1, sample="sample", dryrun="dry"):
+    snakefile = SNAKEFILES[assembler.algorithm]
```
danejo3, line 21: Each assembly will have its' own unique snake file. 

I decided to go with this route because I wanted to each assembly job to have their own DAG. In the Snakemake file, it is possible to go with this route: 

```
if "spades": 
    rule spades:
        blah...

if "megahit":
    rule megahit:
        blah...
```

However, by having these branching statements, there will be a problem of conditional generated DAGs. Maybe it's okay to have this, but I think having a complete DAG job of each assembly would be better than an incomplete DAG with branching complete DAG jobs based on conditions.

yeat/cli/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+        assemblers = config.AssemblyConfiguration.parse_json(fh)
+    for assembler in assemblers:
+        run(
+            *args.reads,
+            assembler=assembler,
+            outdir=args.outdir,
+            cores=args.threads,
+            sample=args.sample,
+            dryrun=args.dry_run,
+        )
```
danejo3, line 58: The purpose of the "for" loop is to run various assemblers that the user requested in the config file.

As of right now, the loop is done iteratively.

Ideally, in the future, being able to run multiple assemblies at the same time would be the better approach when using the grid.

yeat/cli/cli.py
```diff
@@ -7,39 +7,30 @@
...
+                label="assembly1",
+                algorithm="spades",
+            ),
+            dict(
+                label="assembly2",
+                algorithm="megahit",
+            ),
+        ]
+        json.dump(config, sys.stdout, indent=4)
+        raise SystemExit()
```
danejo3, line 29: `yeat --init` will produce a dummy config example printed in the terminal.

yeat/cli/cli.py
```diff
@@ -70,18 +61,16 @@ def get_parser():
         action="store_true",
         help="construct workflow DAG and print a summary but do not execute",
     )
+    parser.add_argument(
+        "config",
+        type=str,
+        help="genome assembly configuration in JSON format; template can be generated by running `yeat --init`",
+    )
```
danejo3, line 68: YEAT will now require users to pass in a "config" file.

danejo3:
Test suite is at 100% coverage.

yeat/cli/__init__.py
danejo3: image

standage (commented):
I think it would he helpful to pull back and look at the big picture here. You’ve raised important software design considerations involving workflow definition and configuration. But these considerations are secondary to *what we’re trying to accomplish with this update.* What is this PR’s primary purpose? A lot of the code and commentary is focused on “implement config file” but the main purpose is “support multiple assemblers,” or more specifically “support for parallel execution of multiple assemblers.” Articulating that purpose clearly in the PR description helps orient the reviewer to effectively evaluate and provide feedback for any proposed changes. But it’s even more important to keep the primary purpose in mind when planning and implementing the changes.

Looking at the code with that in mind, I have a couple concerns.

First, creating a distinct Snakefile for each assembler is incompatible with your stated goal of “in the future, being able to run multiple assemblies at the same time.” You can’t take advantage of all of Snakemake’s nice features supporting parallel execution on and off the grid if you’re invoking each Snakefile separately and sequentially. If splitting up the assembly Snakefile is important from a *design* perspective, you could look into [Snakemake includes](https://snakemake.readthedocs.io/en/stable/snakefiles/modularization.html#includes). (Snakemake also supports more advanced modularization features, but I’d suggest keeping it simple for now). But don’t let secondary *design* considerations take precedence over primary *feature* considerations.

Second, the config file parsing you have implemented here is, frankly, overkill. I agree that it shouldn’t be necessary or permitted for the user to provide duplicate configurations for any assembler. But that also means that the “label” attribute doesn’t serve any purpose and isn’t required. So that leaves us with…a list of assembly algorithm names. It’s overkill to implement a dedicated class, with functions for writing to and reading from JSON, for what amounts to a set of strings—especially when at the moment there’s only one valid string. Don’t get me wrong: I expect that config files will probably be necessary at some point to support configuration of multiple assemblers. And kudos to you for keeping future developments in mind. But you're best served here by waiting to implement configuration until you have something to configure. In the mean time, there are simpler ways to provide a list of assemblers.

standage:
I presume we can close this thread now that #16 has been merged?

danejo3:
> I presume we can close this thread now that #16 has been merged?

Yup! No need for this anymore!

YEAT'd and DELETED!

standage:
> YEAT'd and DELETED!

yeat-delete

PR #13, opened 2022-04-01
Implementing MEGAHIT Into YEAT

### LINKED ISSUE #10: Adding support for MEGAHIT

danejo3:
The purpose of this PR is to close #10 .

In order to run YEAT, users will be required to provide the type of assembler that the user would like to run.

For example,
> yeat --assembly spades [read1] [read2]

In addition to the support for MEGAHIT, if the user would like to run multiple assemblers with the same reads, users can put a comma (",") after each assembler name.

For example,
> yeat --assembly spades,megahit [read1] [read2]

As of right now, YEAT can only assembler read with `spades` and `megahit`. More assemblers will be added in the future.

### DIFF
--- a/yeat/Snakefile
+++ b/yeat/Snakefile
@@ -13,7 +13,8 @@ import pandas as pd
 
 rule all:
     input:
-        expand("analysis/quast/{sample}/report.html", sample=config["sample"]),
+        if config["algorithm"] == "spades":
+            expand("analysis/quast/{sample}/report.html", sample=config["sample"]),
         expand("seq/fastqc/{sample}_{reads}_fastqc.html", sample=config["sample"], reads=["R1", "R2"])
 
 
@@ -104,9 +105,27 @@ rule spades:
     shell: "spades.py -1 {input.read1} -2 {input.read2} -o analysis/spades/{wildcards.sample}"
 
 
-rule quast:
+rule megahit:
+    output:
+        contigs="analysis/megahit/{sample}/final.contigs.fa"
+    input:
+        read1="seq/downsample/{sample}.R1.fq.gz",
+        read2="seq/downsample/{sample}.R2.fq.gz",
+    threads: 8
+    shell: "megahit -1 {input.read1} -2 {input.read2} -o analysis/megahit/{wildcards.sample}"
+
+
+rule spades_quast:
     output:
         report="analysis/quast/{sample}/report.html",
     input:
         scaffold="analysis/spades/{sample}/scaffolds.fasta",
+    shell: "quast.py {input.scaffold} -o analysis/quast/{wildcards.sample}"
+
+
+rule megahit_quast:
+    output:
+        report="analysis/quast/{sample}/report.html",
+    input:
+        scaffold="analysis/megahit/{sample}/scaffolds.fasta",
     shell: "quast.py {input.scaffold} -o analysis/quast/{wildcards.sample}"
--- a/yeat/Snakefile
+++ b/yeat/Snakefile
@@ -13,7 +13,7 @@ import pandas as pd
 
 rule all:
     input:
-        expand("analysis/quast/{assemblers}/{sample}_report.html", assemblers=config["assemblers"], sample=config["sample"]),
+        expand("analysis/quast/{assembler}/{sample}_report.html", assembler=config["assemblers"], sample=config["sample"]),
         expand("seq/fastqc/{sample}_{reads}_fastqc.html", sample=config["sample"], reads=["R1", "R2"])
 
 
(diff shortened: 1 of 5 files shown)


yeat/Snakefile
```diff
@@ -95,18 +95,41 @@ rule downsample:
...
     input:
         read1="seq/downsample/{sample}.R1.fq.gz",
         read2="seq/downsample/{sample}.R2.fq.gz",
     threads: 8
-    shell: "spades.py -1 {input.read1} -2 {input.read2} -o analysis/spades/{wildcards.sample}"
+    shell:
+        """
+        megahit -1 {input.read1} -2 {input.read2} -t {threads} -o analysis/megahit-temp
+        mv analysis/megahit-temp/* analysis/megahit
+        rm -r analysis/megahit-temp
```
danejo3, line 121: MEGAHIT has a weird quirk where it will not run if the output directory already exists. Because of this, when running MEGAHIT, the output is redirected to the `megahit-temp` directory where all of its' contents will be copied over to the `megahit` directory. After all the files are copied over, the `megahit-temp` directory will be deleted.

yeat/cli.py
```diff
@@ -78,8 +105,11 @@ def main(args=None):
     if args is None:
         args = get_parser().parse_args()
     assert len(args.reads) == 2
+    assembly = list(filter(None, args.assembly.strip().split(",")))
```
danejo3, line 108: Takes the string passed in with `--assembly` and splits the substrings by commas. The `filter()` removes any empty strings from the list of strings.

danejo3:
PR is ready for review! @standage

yeat/cli.py
standage: Clever! You may consider renaming this variable from `assembly` to `assemblers` for sake of clarity. A name like "assembly" could have any number of meanings, but "assemblers" is pretty clearly a list of one or more assembly tools/programs/algorithms.

yeat/Snakefile
```diff
@@ -95,18 +95,41 @@ rule downsample:
...
+        """
+
+
+rule megahit:
+    output:
+        contigs="analysis/megahit/{sample}_contigs.fasta"
     input:
         read1="seq/downsample/{sample}.R1.fq.gz",
         read2="seq/downsample/{sample}.R2.fq.gz",
     threads: 8
```
standage, line 116: For use on HPC, something more like 64 or 128 would probably be a much more reasonable default here.

standage (commented):
This looks good to me! I've posted a few comments and questions. One more here:

> One caveat with the current "run multiple assembly algorithms" is that the execution of these assemblies will not be done in parallel but linearly.

Really? If you were to run `yeat --assembly spades,megahit` right now on a machine with 64 threads wouldn't Snakemake run them in parallel?

danejo3:
> Really? If you were to run `yeat --assembly spades,megahit` right now on a machine with 64 threads wouldn't Snakemake run them in parallel?

You are completely right! I didn't understand how Snakemake parallelizes multiple jobs until I just ran it.

danejo3:
Code is ready to be reviewed and merged!

yeat/Snakefile
```diff
@@ -13,7 +13,7 @@ import pandas as pd
 
 rule all:
     input:
-        expand("analysis/quast/{sample}/report.html", sample=config["sample"]),
+        expand("analysis/quast/{assemblers}/{sample}_report.html", assemblers=config["assemblers"], sample=config["sample"]),
```
standage, line 16: The first wildcard in this string makes more sense as `{assembler}` rather than `{assemblers}`. It helps to remember that a wildcard a placeholder that can take on a single value at a time. So a single `{assembler}`, populated with values from `config["assemblers"]` (which *is* appropriate in the multiple tense/cardinality).

standage (commented):
Looks good! Just one more semantic nit to pick. :-)

yeat/Snakefile
danejo3: Understood!

standage (approved):

Issue #8, opened 2022-02-02
Test workflow in test suite


standage:
The three tests currently present in the YEAT test suite test the workflow definition (dry run) and various failure modes. None of the tests actually invoke the workflow. The test suite should be updated to include a function that invokes the entire workflow.

Note: even with an small input, the YEAT workflow requires a non-trivial amount of runtime not conducive to running constantly. We probably want to use [test function attributes](https://docs.pytest.org/en/6.2.x/mark.html) to mark functions that require more runtime, and invoke these only when requested with something like `make testlong`. See Mystt Makefile for an example.

PR #16, opened 2022-09-16
Adding multi-assembly config support

danejo3:
The purpose of this PR is to allow users to create their own assembly config file.

An example of the config file can be seen by executing `yeat --init`, which will pull up:
```
[
    {
        "assembler": "spades",
        "extra_args": ""
    },
    {
        "assembler": "megahit",
        "extra_args": ""
    }
]
```

### DIFF
--- a/yeat/Snakefile
+++ b/yeat/Snakefile
@@ -102,7 +102,7 @@ rule spades:
     threads: 128
     shell:
         """
-        spades.py -1 {input.read1} -2 {input.read2} -t {threads} -o analysis/spades
+        spades.py -1 {input.read1} -2 {input.read2} -t {threads} -o analysis/spades {config[extraargs][spades]}
         ln -s scaffolds.fasta {output.contigs}
         """
 
@@ -116,7 +116,7 @@ rule megahit:
     threads: 128
     shell:
         """
-        megahit -1 {input.read1} -2 {input.read2} -t {threads} -o analysis/megahit-temp
+        megahit -1 {input.read1} -2 {input.read2} -t {threads} -o analysis/megahit-temp {config[extraargs][megahit]}
         mv analysis/megahit-temp/* analysis/megahit
         rm -r analysis/megahit-temp
         ln -s final.contigs.fa {output.contigs}
@@ -131,7 +131,7 @@ rule unicycler:
         read2="seq/downsample/{sample}.R2.fq.gz"
     shell:
         """
-        unicycler -1 {input.read1} -2 {input.read2} -t {threads} -o analysis/unicycler
+        unicycler -1 {input.read1} -2 {input.read2} -t {threads} -o analysis/unicycler {config[extraargs][unicycler]}
         ln -s assembly.fasta {output.contigs}
         """
 
--- a/yeat/Snakefile
+++ b/yeat/Snakefile
@@ -13,7 +13,7 @@ import pandas as pd
 
 rule all:
     input:
-        expand("analysis/quast/{algorithm}/{sample}_report.html", algorithm=config["algorithms"], sample=config["sample"]),
+        expand("analysis/quast/{assembler}/{sample}_report.html", assembler=config["assemblers"], sample=config["sample"]),
         expand("seq/fastqc/{sample}_{reads}_fastqc.html", sample=config["sample"], reads=["R1", "R2"])
 
 
@@ -138,11 +138,11 @@ rule unicycler:
 
 rule quast:
     output:
-        report="analysis/quast/{algorithm}/{sample}_report.html"
+        report="analysis/quast/{assembler}/{sample}_report.html"
     input:
-        contigs="analysis/{algorithm}/{sample}_contigs.fasta"
+        contigs="analysis/{assembler}/{sample}_contigs.fasta"
     shell:
         """
-        quast.py {input.contigs} -o analysis/quast/{wildcards.algorithm}
+        quast.py {input.contigs} -o analysis/quast/{wildcards.assembler}
         ln -s report.html {output.report}
         """
--- a/yeat/Snakefile
+++ b/yeat/Snakefile
@@ -13,7 +13,7 @@ import pandas as pd
 
 rule all:
     input:
-        expand("analysis/quast/{assembler}/{sample}_report.html", assembler=config["assemblers"], sample=config["sample"]),
+        expand("analysis/quast/{algorithm}/{sample}_report.html", algorithm=config["algorithms"], sample=config["sample"]),
         expand("seq/fastqc/{sample}_{reads}_fastqc.html", sample=config["sample"], reads=["R1", "R2"])
 
 
@@ -138,11 +138,11 @@ rule unicycler:
 
 rule quast:
     output:
-        report="analysis/quast/{assembler}/{sample}_report.html"
+        report="analysis/quast/{algorithm}/{sample}_report.html"
     input:
-        contigs="analysis/{assembler}/{sample}_contigs.fasta"
+        contigs="analysis/{algorithm}/{sample}_contigs.fasta"
     shell:
         """
-        quast.py {input.contigs} -o analysis/quast/{wildcards.assembler}
+        quast.py {input.contigs} -o analysis/quast/{wildcards.algorithm}
         ln -s report.html {output.report}
         """
--- a/yeat/Snakefile
+++ b/yeat/Snakefile
@@ -100,9 +100,11 @@ rule spades:
         read1="seq/downsample/{sample}.R1.fq.gz",
         read2="seq/downsample/{sample}.R2.fq.gz"
     threads: 128
+    params:
+        extra_args=config["extra_args"]
     shell:
         """
-        spades.py -1 {input.read1} -2 {input.read2} -t {threads} -o analysis/spades {config[extraargs][spades]}
+        spades.py -1 {input.read1} -2 {input.read2} -t {threads} -o analysis/spades {params.extra_args[spades]}
         ln -s scaffolds.fasta {output.contigs}
         """
 
@@ -114,9 +116,11 @@ rule megahit:
         read1="seq/downsample/{sample}.R1.fq.gz",
         read2="seq/downsample/{sample}.R2.fq.gz"
     threads: 128
+    params:
+        extra_args=config["extra_args"]
     shell:
         """
-        megahit -1 {input.read1} -2 {input.read2} -t {threads} -o analysis/megahit-temp {config[extraargs][megahit]}
+        megahit -1 {input.read1} -2 {input.read2} -t {threads} -o analysis/megahit-temp {params.extra_args[megahit]}
         mv analysis/megahit-temp/* analysis/megahit
         rm -r analysis/megahit-temp
         ln -s final.contigs.fa {output.contigs}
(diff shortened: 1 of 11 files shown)


danejo3:
Tests are passing. Coverage is at 99%. Running full test suite is a bit slow because two tests are running `spades`, `megabit`, and `unicycler`. Will work on #8 to resolve this issue.

danejo3:
This PR is ready for review!

-Create a `config.py` class to parse through the user's config file.
-Added a command to display a template config file
-No changes to snake file was made; will be updating it in the next PR to resolve several issue threads like #17 and #18 .

yeat/cli.py
```diff
@@ -8,35 +8,43 @@
...
-        if algorithm in algorithms:
-            message = (
-                f"Found duplicate assembly algorithm with `--assemblers` flag: [[{algorithm}]]!"
-            )
-            raise ValueError(message)
-        algorithms.append(algorithm)
+class InitAction(Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        json.dump(CONFIG_TEMPLATE, sys.stdout, indent=4)
+        print()
```
danejo3, line 36: Added this for my own OCD sanity. Without this, the terminal would display the template string without a newline character for the terminal command.

yeat/cli.py
```diff
@@ -8,35 +8,43 @@
...
+CONFIG_TEMPLATE = [
+    dict(
+        assembler="spades",
+        extra_args="",
+    ),
+    dict(
+        assembler="megahit",
+        extra_args="",
+    ),
+]
```
standage, line 30: Since this is for demo purposes, it would be appropriate to populate the `extra_args` argument. Maybe something like this.

```python
CONFIG_TEMPLATE = [
     dict(
         assembler="spades",
         extra_args="--meta",
     ),
     dict(
         assembler="megahit",
         extra_args="--min-count 5 --min-contig-len 300",
     ),
 ]
```

standage (changes requested):
In addition to my PR and comment about the mock config, I'll also note that it doesn't appear the extra arguments are actually passed to the Snakefile and used when running the assembly tools.

danejo3:
Thanks for the comments and suggestions in #19 ! Changes and suggestions were incorporated. All tests passing!

standage:
> I'll also note that it doesn't appear the extra arguments are actually passed to the Snakefile and used when running the assembly tools.

Has this been resolved? Could you point me to where the extra arguments are used when running the various assemblers?

danejo3:
> It looks like you didn't merge #19 into this branch. Did you copy and paste the suggested changes? Or did you adapt them? It's hard to tell.

I did not merge #19 original, but copied pasted. After you mentioned the word `merging`, I realized that I could have done just that! 😅 Did I merge the `suggest/config` branch correctly? Hopefully the git commands that I ran were correct.

danejo3:
> > I'll also note that it doesn't appear the extra arguments are actually passed to the Snakefile and used when running the assembly tools.
> 
> Has this been resolved? Could you point me to where the extra arguments are used when running the various assemblers?

In the PR's description, I have mentioned that the `extra_args` value in the config file has no functionality. I was planning on opening up a new MR to resolve the non-functional feature. I did this because I wanted to keep this PR relatively small. However, since the `extra_args` data in the config file has been added, I guess the idea that `extra_args` is functional should be added into this branch. (I probably should have made a different branch with the `extra_args` and its' functionaility.) Anyways, I am currently adapting the snakelike file to execute the extra args for the various assembly algorithms. I'll have it in the get push.

standage:
> Did I merge the `suggest/config` branch correctly?

The idea of opening PR #19, like all PRs, was to have you review it, respond if needed, and merge it. The only difference, as mentioned, is that #19 updated *this* branch instead of `main`. It sounds like you took a different route and merged it on the command line? In any case, whatever you did seems to have integrated my suggestions and closed #19.

standage:
> I was planning on opening up a new MR to resolve the non-functional feature.

Ah I see, I misunderstood that.

I speak often about managing the scope of each PR so that effective review can be done with a feasible amount of time and effort. Was the idea here to split the changes up so that they're more manageable? If so, I'd recommend that you go ahead and include that feature: it's within scope of this branch; this branch isn't terribly large; only marginal additional changes to the code are required; and merging a partially implemented feature is something we typically want to avoid unless it really can't be helped.

If that wasn't the idea, please clarify.

danejo3:
> > Did I merge the `suggest/config` branch correctly?
> 
> The idea of opening PR #19, like all PRs, was to have you review it, respond if needed, and merge it. The only difference, as mentioned, is that #19 updated _this_ branch instead of `main`. It sounds like you took a different route and merged it on the command line? In any case, whatever you did seems to have integrated my suggestions and closed #19.

Oh, I see. Yes, I was under the impression that I needed to merge it back into my branch.

danejo3:
> Was the idea here to split the changes up so that they're more manageable? 

Yeah, I'm trying to break down my tasks to more manageable chucks for people to follow.

> If so, I'd recommend that you go ahead and include that feature: it's within scope of this branch; this branch isn't terribly large; only marginal additional changes to the code are required; and merging a partially implemented feature is something we typically want to avoid unless it really can't be helped.

Sounds good. Updating the code as we speak!

standage:
> I was under the impression that I needed to merge it back into my branch.

The big green [Merge] button on the PR is provided conveniently for that purpose. :smiley:

This is a less common use case, but it can be helpful when proposing changes to an existing branch/PR. In the future, if you approve of the suggested changes, you can just approve and merge the PR and it will update your existing branch/PR. No command line drudgery required.

yeat/Snakefile
```diff
@@ -102,7 +102,7 @@ rule spades:
     threads: 128
     shell:
         """
-        spades.py -1 {input.read1} -2 {input.read2} -t {threads} -o analysis/spades
+        spades.py -1 {input.read1} -2 {input.read2} -t {threads} -o analysis/spades {config[extraargs][spades]}
```
danejo3, line 105: In the snakemake file, we then call the dictionary and retrieve the key's value based on assembly.

Because we do a hard check checking for valid assembly algorithms in the config file, I hardcoded the keys here for each assembly algorithm.

yeat/cli.py
```diff
@@ -8,39 +8,54 @@
...
+    extraargs = {}
+    for algorithm in algorithms:
+        assembler = algorithm.algorithm
+        assemblers.append(assembler)
+        extraargs[assembler] = algorithm.extra_args
     config = dict(
         read1=r1,
         read2=r2,
         assemblers=assemblers,
+        extraargs=extraargs,
```
danejo3, line 58: Changed variable names for improved readability.

yeat/Snakefile
standage: You're on the right path here, and the hard-coding is appropriate. Honestly, this is a perfectly viable solution.

The only suggestion I would offer involves Snakemake `params`. Sometimes when bracketed values in my `shell` statements get too hairy, I pull that out into the rule's `params` block. That gives more flexibility (especially if function calls are needed) and keeps the `shell` statement clean. For this rule, it might look something like this.

```python
rule spades:
    output:
        contigs="analysis/spades/{sample}_contigs.fasta"
    input:
        read1="seq/downsample/{sample}.R1.fq.gz",
        read2="seq/downsample/{sample}.R2.fq.gz"
    threads: 128
    params:
        extra_args=config["extra_args"]["spades"]
    shell:
        """
        spades.py -1 {input.read1} -2 {input.read2} -t {threads} -o analysis/spades {params.extra_args}
        ln -s scaffolds.fasta {output.contigs}
        """
```

yeat/cli.py
```diff
@@ -8,39 +8,54 @@
...
-        algorithms.append(algorithm)
+class InitAction(Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        json.dump(CONFIG_TEMPLATE, sys.stdout, indent=4)
+        print()
+        raise SystemExit()
 
 
-def run(read1, read2, assemblers, outdir=".", cores=1, sample="sample", dryrun="dry"):
+def run(read1, read2, algorithms, outdir=".", cores=1, sample="sample", dryrun="dry"):
```
standage, line 40: Hey man, I feel this. We joke often that naming things is one of the hardest problems in computer science, but as funny as it is it's true!

Both `assemblers` and `algorithms` are reasonable variable names here. It's a collection of `AssemblerConfig` objects, each of which has an `algorithm` attribute. So `configs` or `assembly_configs` would also be appropriate. Consider the suggested code in the comment below.

I'm definitely well into hair splitting territory by now, but `fastq{1,2}` might also be more appropriate names for the first two arguments than `read{1,2}`.

yeat/cli.py
```diff
@@ -8,39 +8,54 @@
...
+        raise FileNotFoundError(f"No such file: '{r1}'")
     r2 = Path(read2).resolve()
+    if not r2.is_file():
+        raise FileNotFoundError(f"No such file: '{r2}'")
+    assemblers = []
+    extraargs = {}
+    for algorithm in algorithms:
+        assembler = algorithm.algorithm
+        assemblers.append(assembler)
+        extraargs[assembler] = algorithm.extra_args
```
standage, line 53: For what it's worth, these six lines could be collapsed into two arguably clearer lines using list and dict comprehensions.

```python
assemblers = [config.algorithm for config in assembly_configs]
extra_args = {config.algorithm: config.extra_args for config in assembly_configs}
```

standage (commented):
Looking good! A couple a minor comments for your consideration.

yeat/Snakefile
```diff
@@ -100,9 +100,11 @@ rule spades:
         read1="seq/downsample/{sample}.R1.fq.gz",
         read2="seq/downsample/{sample}.R2.fq.gz"
     threads: 128
+    params:
+        extra_args=config["extra_args"]
     shell:
         """
-        spades.py -1 {input.read1} -2 {input.read2} -t {threads} -o analysis/spades
+        spades.py -1 {input.read1} -2 {input.read2} -t {threads} -o analysis/spades {params.extra_args[spades]}
```
danejo3, line 107: Tried the:

>params:
         extra_args=config["extra_args"]["spades"]

The code works; however, after messing with the code for a while, I came to learn that all parameters in snakemake rules are initialized. Because of this, if the config file did not have a certain assembly algorithm, for example, `spades`, would fail on the `config["extra_args"]["spades"]` because the key does not exist. To prevent that, there is a few solutions that I could think of.

1) You can encapsulate the entire rule and only run if the key exists (but doing this option will open a whole new set of problems). Or,
2) Get the value from the dictionary only when the shell command is executed (meaning that the user's config has the desired assembly algorithm to run).

In the end, I went with the latter and learned that if `config["extra_args"]` does not have the key, snakemake will crash and fail. As a result, the following code access the key's value must be down in the shell command.

standage: Ah, good catch. So my suggested change doesn't make the code as concise as I thought. Thanks for the clear explanation. I think this is a good compromise.

standage (approved):

danejo3:
Thanks for all of your suggestions!

yeat/Snakefile
danejo3: You more you know!

PR #20, opened 2022-09-22
Run long tests when needed

danejo3:
First commit

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,14 @@ help: Makefile
 
 ## test:        run automated test suite
 test:
+	pytest --cov=yeat -m 'not long'
+
+## testlong:    run only long-running automated tests
+testlong:
+	pytest --cov=yeat -m long
+
+## testall:     run all tests
+testall:
 	pytest --cov=yeat
 
 ## style:       check code style vs Black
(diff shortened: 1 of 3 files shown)


danejo3:
PR should be relatively straight forward and will help with test suite time management. Let me know if there is anything else that I need to add or change. Thanks!

Makefile
```diff
@@ -9,6 +9,14 @@ help: Makefile
 
 ## test:        run automated test suite
 test:
+	pytest --cov=yeat -m 'not long'
+
+## testlong:    run only long-running automated tests
+testlong:
+	pytest --cov=yeat -m long
```
danejo3, line 16: Quick question about `coverage`. When you break up tests, would be best to for go `coverage` since you will not be running most of the codebase? Would it make more sense to only test for coverage in `make testall`.

Currently `make testall` is at 99% coverage; however, others have lower amounts. I can see where showing coverage can still be helpful, but I feel like it's not helpful at the sometime. 🤷‍♂️

standage: Yeah, measuring coverage on a subset of tests doesn't give you very much insight.

standage (approved):
Looks good to me!

PR #21, opened 2022-09-23
Adding custom downsample flag

### LINKED ISSUE #17: Add optional custom downsample number
Currently, YEAT has its' own formula for calculating the value to downsample all given reads before running an assembly algorithm.

Here is the formula:
> down = (genome_size * 150) // (2 * read_length)

Sometimes, the downsample value calculated by YEAT may not be the best number. As a result, it would be nice to have an option where users can input a custom downsample number to override the default calculation.

A request for this feature was made a long time ago, but I do not recall the situation that called for this idea.

danejo3:
The purpose of this PR is to resolve #17 .

Originally, YEAT would automatically downsample to a x150 coverage. In this PR, a new option was added for users to input their own custom downsampling number using the `-d` flag.

From my understanding, depending on certain situations, not all reads should conform to an estimated downsample number because you might need more or less total reads.

### DIFF
--- a/yeat/Snakefile
+++ b/yeat/Snakefile
@@ -79,15 +79,21 @@ rule downsample:
         read1="seq/fastp/{sample}.R1.fq.gz",
         read2="seq/fastp/{sample}.R2.fq.gz",
         mash_report="seq/mash/{sample}.report.tsv"
+    params:
+        downsample=config["downsample"]
     run:
         result = input.mash_report
         df = pd.read_csv(result, sep="\t")
         genome_size = df["Length"].iloc[0]
-        coverage = 150
-        read_length = 250
-        down = (genome_size * 150) // (2 * read_length)
         print(f"genome size: {genome_size}")
-        print(f"down: {down}")
+        if params.downsample == 0:
+            coverage = 150
+            read_length = 250
+            down = (genome_size * coverage) // (2 * read_length)
+            print(f"auto down: {down}")
+        else:
+            down = params.downsample
+            print(f"custom down: {down}")
         shell("seqtk sample {input.read1} {down} > seq/downsample/{wildcards.sample}.R1.fq")
         shell("seqtk sample {input.read2} {down} > seq/downsample/{wildcards.sample}.R2.fq")
         shell("gzip -f seq/downsample/*")
--- a/yeat/Snakefile
+++ b/yeat/Snakefile
@@ -9,6 +9,7 @@
 
 from shutil import copyfile
 import pandas as pd
+from pathlib import Path
 
 
 rule all:
@@ -82,10 +83,16 @@ rule downsample:
     params:
         downsample=config["downsample"]
     run:
+        if params.downsample == -1:
+            p = Path("seq/downsample")
+            p.mkdir(parents=True, exist_ok=True)
+            copyfile(input.read1, output.sub1)
+            copyfile(input.read2, output.sub2)
+            return
         result = input.mash_report
         df = pd.read_csv(result, sep="\t")
         genome_size = df["Length"].iloc[0]
-        print(f"genome size: {genome_size}")
+        print(f"genome size: {genome_size}")    
         if params.downsample == 0:
             coverage = 150
             read_length = 250
--- a/yeat/cli.py
+++ b/yeat/cli.py
@@ -117,7 +117,7 @@ def get_parser():
         type=int,
         metavar="D",
         default=0,
-        help="downsample reads down to a custom number; by default, D=0; when no value is given, YEAT will auto-downsample",
+        help="downsample reads down to a custom number; by default, D=0; when set to default, YEAT will auto downsample; set D=-1 to not downsample",
     )
     parser.add_argument(
         "--init",
--- a/yeat/cli.py
+++ b/yeat/cli.py
@@ -117,7 +117,7 @@ def get_parser():
         type=int,
         metavar="D",
         default=0,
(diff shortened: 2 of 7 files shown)


danejo3:
Question: Is there ever a time when you never what to downsample?

I just talked with Diana and she said that, yes, when you are doing metagenomic analysis, you never want to downsample because you want all the reads.

danejo3:
> From my understanding, depending on certain situations, not all reads should conform to an estimated downsample number because you might need more or less total reads.

From Sean: It comes up the most when we sequence a plasmid. We wind up with 100,000x+ coverage and the way mash calculates genome size breaks down.

standage:
> ...when you are doing metagenomic analysis, you never want to downsample because you need all the reads.

Yep. One of the problems with shotgun metagenomics is that taxa are present in the sample at uneven abundances. Generating more reads can only help so much, because sequencing more deeply is going to give you piles of reads from the high abundance taxa that are already well covered, and only a few reads of interest from the low abundance taxa.

I *will* note that the complexity and scale of some metagenomics samples makes assembly difficult or impossible without subsampling. But in this case, you wouldn't just perform random uniform sampling: you'd apply a *digital normalization* approach that reduces the coverage of highly abundant sequences but keeps all sequences of low abundance. The result is a drastic reduction in *data* (often up to 90%) with very little impact on *information content* (e.g. which *k*-mers are present to populate the assembly graph). But this is only true for metagenomic *assembly*: for metagenomic *profiling*, diginorm would indeed be inappropriate.

yeat/Snakefile
```diff
@@ -79,15 +80,27 @@ rule downsample:
...
         mash_report="seq/mash/{sample}.report.tsv"
+    params:
+        downsample=config["downsample"]
     run:
+        if params.downsample == -1:
+            p = Path("seq/downsample")
+            p.mkdir(parents=True, exist_ok=True)
+            copyfile(input.read1, output.sub1)
+            copyfile(input.read2, output.sub2)
+            return
```
danejo3, line 91: Learned that you actually can `symlink` instead of copy files; however, the current snakemake version is unable to do that until 3.8. (We are currently on 3.7.)

```
Path(output.sub1).symlink_to(Path(input.read1))
Path(output.sub2).symlink_to(Path(input.read2)) 
```

image

standage: Don't confuse Python 3.7 with Snakemake 3.7. We support Python 3.7 for many of our workflows, but Snakemake 3.7 is *ancient* in bioinformatics years. :smiley: At the moment I have `snakemake>=6.0,<7.0` pinned on several repos, since there seem to be some error handling regressions with Snakemake version 7.

yeat/tests/test_cli.py
```diff
@@ -92,3 +93,30 @@ def test_unicycler(capsys, tmp_path):
     cli.main(args)
     assembly_result = Path(wd).resolve() / "analysis" / "unicycler" / "assembly.fasta"
     assert assembly_result.exists()
+
+
+@pytest.mark.long
+@pytest.mark.parametrize(
+    "downsample,num_contigs,largest_contig,total_len",
+    [("2000", 71, 5120, 69189), ("-1", 56, 35168, 199940)],
```
danejo3, line 101: Tests looks at the quast `report.tsv` file when comparing the expected output.

danejo3:
I think this PR is ready for a review! Adding @lovettse if you have any comments or questions. Thanks!

yeat/cli.py
```diff
@@ -101,6 +111,14 @@ def get_parser():
...
         help="construct workflow DAG and print a summary but do not execute",
     )
+    parser.add_argument(
+        "-d",
+        "--downsample",
+        type=int,
+        metavar="D",
+        default=0,
+        help="downsample reads down to the desired number; by default, D=0; when set to default, YEAT will auto downsample; set D=-1 to not downsample",
+    )
```
standage, line 121: Cool. Three modes:

1. Auto-downsample to 150x coverage (D=0)
2. Downsample to a fixed number of reads D (D>0)
3. Disable downsampling (D<0)

yeat/Snakefile
```diff
@@ -79,15 +80,27 @@ rule downsample:
...
-        coverage = 150
-        read_length = 250
-        down = (genome_size * 150) // (2 * read_length)
-        print(f"genome size: {genome_size}")
-        print(f"down: {down}")
+        print(f"genome size: {genome_size}")    
+        if params.downsample == 0:
+            coverage = 150
+            read_length = 250
+            down = (genome_size * coverage) // (2 * read_length)
```
standage, line 99: You may want to make 150 a parameter that the user can configure with the CLI.

standage: Also, read length for autosampling is...tricky. If they're untrimmed 250 bp reads that's great, but if they're not 250 bp read or if they're trimmed, you'd need to determine the distribution of the read lengths and use the mean or median of that distribution to determine the *approximate* number of reads needed for the desired coverage. A little more involved, and something that might benefit from a dedicated branch/thread.

standage (approved):
Looks good to me, but I'll give Sean a chance to respond before merging.

yeat/Snakefile
lovettse: This has always bothered me a bit, too. We don't need a ton of precision, so it's OK to just land in the right ballpark (125x tends to be good enough if I asked for 150x). That said, we just had a run where the fragment size was so small that a 2x250 run was essentially 2x150 after trimming, and that's not all that uncommon. In the past, we've handled it by just asking for more coverage when we knew the fragment sizes were small, but that's not a terribly satisfying solution.

PR #23, opened 2022-10-19
Improved auto downsampling with custom coverage and avg read length

### LINKED ISSUE #17: Add optional custom downsample number

danejo3:
The purpose of this PR is to resolve #17 and improve the `downsample` rule even further in #21 .

### DIFF
--- a/yeat/Snakefile
+++ b/yeat/Snakefile
@@ -7,6 +7,7 @@
 # Development Center.
 # -------------------------------------------------------------------------------------------------
 
+import json
 from shutil import copyfile
 import pandas as pd
 from pathlib import Path
@@ -81,7 +82,9 @@ rule downsample:
         read2="seq/fastp/{sample}.R2.fq.gz",
         mash_report="seq/mash/{sample}.report.tsv"
     params:
-        downsample=config["downsample"]
+        coverage=config["coverage"],
+        downsample=config["downsample"],
+        fastp_report="seq/fastp/fastp.json"
     run:
         if params.downsample == -1:
             p = Path("seq/downsample")
@@ -89,14 +92,21 @@ rule downsample:
             copyfile(input.read1, output.sub1)
             copyfile(input.read2, output.sub2)
             return
-        result = input.mash_report
-        df = pd.read_csv(result, sep="\t")
+        df = pd.read_csv(input.mash_report, sep="\t")
         genome_size = df["Length"].iloc[0]
-        print(f"genome size: {genome_size}")    
+        print(f"genome size: {genome_size}")
+        data = json.load(open(params.fastp_report))
+        read1_length = data["summary"]["after_filtering"]["read1_mean_length"]
+        read2_length = data["summary"]["after_filtering"]["read2_mean_length"]
+        read_lengths = [read1_length, read2_length]
+        avg_length = sum(read_lengths) // len(read_lengths)
+        print(f"average read length: {avg_length}")
+        if params.coverage == 150:
+            print(f"default coverage: {params.coverage}")
+        else:
+            print(f"custom coverage: {params.coverage}")
         if params.downsample == 0:
-            coverage = 150
-            read_length = 250
-            down = (genome_size * coverage) // (2 * read_length)
+            down = (genome_size * params.coverage) // (2 * avg_length)
             print(f"auto down: {down}")
         else:
             down = params.downsample
--- a/yeat/cli.py
+++ b/yeat/cli.py
@@ -46,6 +46,7 @@ def run(
     sample="sample",
     dryrun="dry",
     downsample=0,
+    coverage=0,
 ):
     snakefile = resource_filename("yeat", "Snakefile")
     r1 = Path(fastq1).resolve()
@@ -66,6 +67,7 @@ def run(
         sample=sample,
         dryrun=dryrun,
         downsample=downsample,
+        coverage=coverage,
     )
     success = snakemake(
         snakefile,
@@ -120,6 +122,14 @@ def get_parser():
         help="downsample reads down to the desired number; by default, D=0; when set to default, YEAT will auto downsample; set D=-1 to not downsample",
     )
     parser.add_argument(
+        "-c",
+        "--coverage",
+        type=int,
+        metavar="C",
+        default=150,
+        help="set coverage to the desired number; by default, C=150",
+    )
+    parser.add_argument(
         "--init",
         action=InitAction,
         nargs=0,
@@ -142,4 +152,5 @@ def main(args=None):
         sample=args.sample,
         dryrun=args.dry_run,
         downsample=args.downsample,
+        coverage=args.coverage,
     )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for Unicycler (#14)
 - Support for assembly configuration file (#16)
 - Commands to run a short or long test suite (#20)
-- Custom downsampling flag (#21)
+- Custom downsampling flag (#21, )
 
 
 ## [0.1] 2021-12-01


yeat/Snakefile
```diff
@@ -81,22 +82,31 @@ rule downsample:
...
-        df = pd.read_csv(result, sep="\t")
+        df = pd.read_csv(input.mash_report, sep="\t")
         genome_size = df["Length"].iloc[0]
-        print(f"genome size: {genome_size}")    
+        print(f"genome size: {genome_size}")
+        data = json.load(open(params.fastp_report))
+        read1_length = data["summary"]["after_filtering"]["read1_mean_length"]
+        read2_length = data["summary"]["after_filtering"]["read2_mean_length"]
+        read_lengths = [read1_length, read2_length]
+        avg_length = sum(read_lengths) // len(read_lengths)
```
danejo3, line 102: `avg_length` for all fastq files is compiled here using `fastp`'s report.

danejo3:
PR is ready for review. New coverage flag was added and no more hard-coded read lengths. Let me know if you have any questions or concerns. Thanks!

yeat/Snakefile
standage: Here, you're taking a mean of means: the average of R1 average and R2 average. I'd suggest computing the average length directly from the number of bases and reads.

```python
with open(params.fastp_report, "r") as fh:
    qcdata = json.load(fh)
base_count = qcdata["summary"]["after_filtering"]["total_bases"]
read_count = qcdata["summary"]["after_filtering"]["total_reads"]
avg_read_length = base_count / read_count
```

yeat/cli.py
```diff
@@ -79,6 +81,16 @@ def run(
         raise RuntimeError("Snakemake Failed")
 
 
+def check_positive(value):
+    try:
+        value = int(value)
+        if value <= 0:
+            raise argparse.ArgumentTypeError(f"{value} is not a positive integer")
+    except ValueError:
+        raise Exception(f"{value} is not an integer")
```
standage, line 90: Using `ValueError` here would be more appropriate than `Exception`.

yeat/tests/test_cli.py
```diff
@@ -41,7 +42,7 @@ def test_basic_dry_run(tmp_path):
 
 
 def test_no_args():
-    with pytest.raises(SystemExit, match=r"2"):
+    with pytest.raises(Exception, match=r"yeat is not an integer"):
```
standage, line 45: This test is unclear. Where is `"yeat"` being passed to an argument where an integer is expected? How is this behavior relevant to `yeat` being run with no arguments?

yeat/tests/test_cli.py
```diff
@@ -120,3 +121,56 @@ def test_custom_downsample_input(
...
+        data_file("megahit.cfg"),
+        data_file("short_reads_1.fastq.gz"),
+        data_file("short_reads_2.fastq.gz"),
+        "--coverage",
+        coverage,
+    ]
+    with pytest.raises(SystemExit) as e:
+        args = cli.get_parser().parse_args(arglist)
+    assert isinstance(e.value.__context__, argparse.ArgumentError)
+    assert f"{coverage} is not a positive integer" in e.value.__context__.message
```
standage, line 138: I'm curious how you figured this one out. Any relevant documentation or StackOverflow posts you could point to? I often have this problem, where a `SystemExit` is raised but I want to test the the presence of a different exception. Using `pytest.raises(ArgumentError, match="is not a positive integer")` doesn't work since it's not the ArgumentError being caught. One solution I've used involves catching the `SystemExit` but then checking the contents of `capsys` for the expected error message. I think I like this approach better, but I'm curious if the pytest folks have a recommended syntax here: `isinstance` and `.__context__` feels very ducktape/WD-40 to me. 😃

yeat/tests/test_cli.py
```diff
@@ -120,3 +121,56 @@ def test_custom_downsample_input(
...
+
+
+@pytest.mark.parametrize("coverage", [("string"), ("3.14")])
+def test_invalid_custom_coverage_2(coverage):
```
standage, line 142: Less of a concern for test function names, but instead of `_1` and `_2` you might consider including something like `_negative` and `_noninteger` in the function name.

yeat/tests/test_cli.py
```diff
@@ -120,3 +121,56 @@ def test_custom_downsample_input(
...
+        coverage,
+    ]
+    with pytest.raises(Exception, match=r"is not an integer"):
+        args = cli.get_parser().parse_args(arglist)
+
+
+@pytest.mark.long
+@pytest.mark.parametrize(
+    "coverage,num_contigs,largest_contig,total_len",
+    [("125", 56, 35168, 199940), ("150", 56, 35168, 199940)],
```
standage, line 157: You define 4 parameters, but 3 are invariant across all tests. It's ok to hard-code parameters that are the same for each test instance. Although, it would be interesting to include a test that resulted in a different number of contigs, total length, etc.

standage (commented):
Looks good overall, but I've included a few comments and questions.

yeat/tests/test_cli.py
danejo3: I got my answer from both the pytest documentation and stack overflow; however, did not find an "official" solution. I've had this problem in the past and I too did the capsys string checking but I was curious if there was a better way and found this.

https://stackoverflow.com/questions/49323496/check-argparse-argumenttypeerror

Here is the the documentation page from pytest for the `ArgumentError` class.

https://docs.pytest.org/en/7.0.x/_modules/_pytest/config/argparsing.html

I'm not sure what is the better way, but it works. 🤔

yeat/tests/test_cli.py
danejo3: Thanks for the suggestion!

yeat/cli.py
```diff
@@ -79,6 +81,16 @@ def run(
         raise RuntimeError("Snakemake Failed")
 
 
+def check_positive(value):
+    try:
+        value = int(value)
+        if value <= 0:
+            raise argparse.ArgumentTypeError(f"{value} is not a positive integer")
+    except ValueError:
+        raise argparse.ArgumentTypeError(f"{value} is not an integer")
```
danejo3, line 90: >Using ValueError here would be more appropriate than Exception.

I tried to raise a `ValueError`; however, every time the exception is thrown, the `argparse.ArgumentTypeError` is thrown instead. It seems like it is because the function, `check_positive` is called in the CLI. I tried a variety of ways to catch the ValueError for testing, but I was unable to figure out how to do so.

yeat/cli.py
```diff
@@ -119,6 +131,14 @@ def get_parser():
         default=0,
         help="downsample reads down to the desired number; by default, D=0; when set to default, YEAT will auto downsample; set D=-1 to not downsample",
     )
+    parser.add_argument(
+        "-c",
+        "--coverage",
+        type=check_positive,
```
danejo3, line 137: This is where the `check_positive` is being called.

To be honest, we can solve the throw `ValueError` problem above if I changed this to an `int` instead of `check_positive`.

In doing so, in order to check if the `--coverage` is <= 0, I would need to check after the CLI is parsed (probably in the snakemake). Instead, I followed the parse_args option by assigning the flag input `type` with a function.

yeat/tests/test_cli.py
danejo3: Reverted back to the original test. The reason for this exception test is because when we call the cli with no arguments, in the command line, `yeat` is considered a CLI argument. I do understand why this is the case.

For example, I were to type `yeat` into the terminal, I would get this error:
>usage: yeat [-h] [-v] [-o DIR] [-t T] [--sample S] [-n] [-d D] [-c C] [--init] config reads reads
yeat: error: the following arguments are required: config, reads

When I run the cli with no input, `cli.main(None)`, I get the following error:
usage: yeat [-h] [-v] [-o DIR] [-t T] [--sample S] [-n] [-d D] [-c C] [--init] config reads reads
yeat: error: the following arguments are required: reads

Notice that `config` is missing the error message above.

The reason why the outdated test had a regex string `yeat is not an integer` is because `yeat` would be considered a string in the check positive function.

I am unclear how this is the case, but this is the way the argparse parses a empty string command.

yeat/tests/test_cli.py
```diff
@@ -120,3 +121,55 @@ def test_custom_downsample_input(
...
+        data_file("megahit.cfg"),
+        data_file("short_reads_1.fastq.gz"),
+        data_file("short_reads_2.fastq.gz"),
+        "--coverage",
+        coverage,
+    ]
+    with pytest.raises(SystemExit) as error:
+        args = cli.get_parser().parse_args(arglist)
+    assert isinstance(error.value.__context__, argparse.ArgumentError)
+    assert f"{coverage} is not an integer" in error.value.__context__.message
```
danejo3, line 153: Like the test above, since I am now raising a `ArgumentError` for "not int" values, I'm doing the same here.

It seems like there are a few comments on the recent question you posted on `pytest`'s GitHub Q&A.

https://github.com/pytest-dev/pytest/discussions/10440

yeat/tests/test_cli.py
danejo3: > Although, it would be interesting to include a test that resulted in a different number of contigs, total length, etc.

Any suggestions on to test this? I completely agree and that is what I was hoping to do in this test; however, every time I tried to change the coverage, I was getting the same values of `num_contigs`, `largest_contig`, and `total_len`. The only time I was able to get something different was when I had set the coverage down to `1` and the number of contigs went down to 56 -> 31. The largest config stayed the same and the total length slightly decreased by a 100 bases.

I ultimately didn't go with 1x coverage because of its unrealistic input value (I could be wrong here).

Any suggestions would be welcomed. As of right now, I just have the different coverages and making sure that the values didn't change.

yeat/tests/test_cli.py
danejo3: From what I'm seeing, the `exit_on_error` flag looks the best.

danejo3: UPDATE: Added the new flag and the tests were consolidated down to a regex match.

yeat/tests/test_cli.py
```diff
@@ -40,11 +41,6 @@ def test_basic_dry_run(tmp_path):
     cli.main(args)
 
 
-def test_no_args():
-    with pytest.raises(SystemExit, match=r"2"):
-        cli.main(None)
```
danejo3, line 45: Removed this test overall. Coverage went down slightly.

This test works fines when ran by itself, but fails when running the test suite.

In order for this test to pass, you'll need to do this:
>def test_no_args():
     with pytest.raises(argparse.ArgumentError, match=r"yeat is not an integer"):
         cli.main(None)

Again, this problem touches the `type=check_positive`.

yeat/tests/test_cli.py
danejo3: UPDATE: Test was removed.

standage (approved):
Looks good!

PR #25, opened 2022-10-27
Suggestions for downsampling

standage:
Rather than commenting directly on #23, I'm placing my comments and suggestions in this PR. If you have any questions or concerns, feel free to discuss them here. Otherwise, you can merge this to update the branch associated with your open PR. Additional comments pending.

### DIFF
--- a/yeat/Snakefile
+++ b/yeat/Snakefile
@@ -11,6 +11,7 @@ import json
 from shutil import copyfile
 import pandas as pd
 from pathlib import Path
+from random import randint
 
 
 rule all:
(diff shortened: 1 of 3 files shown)


yeat/Snakefile
```diff
@@ -100,19 +101,17 @@ rule downsample:
...
-            print(f"custom down: {down}")
-        shell("seqtk sample {input.read1} {down} > seq/downsample/{wildcards.sample}.R1.fq")
-        shell("seqtk sample {input.read2} {down} > seq/downsample/{wildcards.sample}.R2.fq")
+        seed = randint(1, 2**16-1)
+        print(f"[yeat] average read length: {avg_read_length}")
+        print(f"[yeat] target depth of coverage: {params.coverage}x")
+        print(f"[yeat] number of reads to sample: {down}")
+        print(f"[yeat] random seed for sampling: {seed}")
+        shell("seqtk sample -s {seed} {input.read1} {down} > seq/downsample/{wildcards.sample}.R1.fq")
+        shell("seqtk sample -s {seed} {input.read2} {down} > seq/downsample/{wildcards.sample}.R2.fq")
```
standage, line 114: The two `seqtk sample` commands need the same seed so that the "random" subset of reads chosen from each file is the same. But in routine use this should really be a randomly (or at least arbitrarily) chosen number. We'll print it out here for debugging purposes, in case a particular seed results in bad behavior that needs to be reproduced.

yeat/cli.py
```diff
@@ -85,37 +84,37 @@ def check_positive(value):
...
     except ValueError:
-        raise argparse.ArgumentTypeError(f"{value} is not an integer")
+        raise ArgumentTypeError(f"{value} is not an integer")
     return value
 
 
-def get_parser():
-    parser = argparse.ArgumentParser(exit_on_error=False)
+def get_parser(exit_on_error=True):
+    parser = ArgumentParser(exit_on_error=exit_on_error)
```
standage, line 94: Just pushed a new commit with the changes we discussed: here...

yeat/tests/test_cli.py
```diff
@@ -140,8 +140,8 @@ def test_invalid_custom_coverage_noninteger(coverage):
         "--coverage",
         coverage,
     ]
-    with pytest.raises(argparse.ArgumentError, match=rf"{coverage} is not an integer"):
-        args = cli.get_parser().parse_args(arglist)
+    with pytest.raises(ArgumentError, match=rf"{coverage} is not an integer"):
+        args = cli.get_parser(exit_on_error=False).parse_args(arglist)
```
standage, line 144: ...and here.

yeat/Snakefile
danejo3: Good to know!

danejo3:
Thanks for the help and suggestions!

PR #29, opened 2022-11-02
Adding `--seed` flag to customize the `seed` when downsampling

### LINKED ISSUE #27: test_custom_downsample_input() test failing
After the recent PR (#24), the `test_custom_downsample_input()` has failed because of the recent changes made in the snakemake `downsample` rule. In the rule, we added `randint` to randomize the downsample process when running `seqtk`. Because of the random seed will be a different number every time the rule is ran, the `num_contigs` and `largest_contig` can change in the `quast` report.

danejo3:
The purpose of this PR is to resolve #27 by adding a `--seed` flag to ensure that the original test `test_custom_downsample_input()` could reproduce the same number of contigs, config lengths, and total length.

Other additions to this PR include: 
- a small update to the `README.md`, 
- an additional test testing for an acceptable range of expected values because of the random seed that was introduced in #23 , and 
- enforcing users to have python `>v3.9` because of the `exit_on_error` parameter that was introduced in #23 as well.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -4,13 +4,14 @@ channels:
     - bioconda
     - defaults
 dependencies:
-    - black=21.10b0
+    - black=22.10
     - fastp>=0.23
     - fastqc>=0.11
     - gzip>=1.7
     - mash>=2.3
     - megahit>=1.2
     - pytest-cov>=3.0
+    - python>=3.9
     - quast>=5.0
     - seqtk>=1.3
     - snakemake>=6.10
--- a/environment.yml
+++ b/environment.yml
@@ -5,13 +5,14 @@ channels:
     - defaults
 dependencies:
     - black=22.10
+    - click=8.0
     - fastp>=0.23
     - fastqc>=0.11
     - gzip>=1.7
     - mash>=2.3
     - megahit>=1.2
     - pytest-cov>=3.0
-    - python>=3.9
+    - python=3.9
     - quast>=5.0
     - seqtk>=1.3
     - snakemake>=6.10
--- a/environment.yml
+++ b/environment.yml
@@ -5,7 +5,6 @@ channels:
     - defaults
 dependencies:
     - black=22.10
-    - click=8.0
     - fastp>=0.23
     - fastqc>=0.11
     - gzip>=1.7
--- a/yeat/Snakefile
+++ b/yeat/Snakefile
@@ -107,8 +107,10 @@ rule downsample:
             down = int((genome_size * params.coverage) / (2 * avg_read_length))
         else:
             down = params.downsample
-        random.seed(params.seed) # for unit tests
-        seed = randint(1, 2**16-1)
+        if params.seed == None:
+            seed = randint(1, 2**16-1)
+        else:
+            seed = params.seed
         print(f"[yeat] average read length: {avg_read_length}")
         print(f"[yeat] target depth of coverage: {params.coverage}x")
         print(f"[yeat] number of reads to sample: {down}")
--- a/yeat/Snakefile
+++ b/yeat/Snakefile
@@ -11,6 +11,7 @@ import json
 from shutil import copyfile
 import pandas as pd
 from pathlib import Path
+import random
 from random import randint
 
 
@@ -85,7 +86,8 @@ rule downsample:
     params:
         coverage=config["coverage"],
         downsample=config["downsample"],
-        fastp_report="seq/fastp/fastp.json"
+        fastp_report="seq/fastp/fastp.json",
+        seed=config["seed"]
     run:
         if params.downsample == -1:
             p = Path("seq/downsample")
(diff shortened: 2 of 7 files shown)


yeat/Snakefile
```diff
@@ -105,6 +107,7 @@ rule downsample:
             down = int((genome_size * params.coverage) / (2 * avg_read_length))
         else:
             down = params.downsample
+        random.seed(params.seed) # for unit tests
         seed = randint(1, 2**16-1)
```
danejo3, line 111: The `random.seed()` setting is the only solution that I could think of to enforce `randint` to be the same number every time this rule is ran for pytest.

yeat/tests/test_cli.py
```diff
@@ -164,3 +167,25 @@ def test_custom_coverage_input(coverage, capsys, tmp_path):
...
+        "-d",
+        "2000",
+    ]
+    args = cli.get_parser().parse_args(arglist)
+    cli.main(args)
+    quast_report = Path(wd).resolve() / "analysis" / "quast" / "megahit" / "report.tsv"
+    df = pd.read_csv(quast_report, sep="\t")
+    assert 61 <= df.iloc[12]["sample_contigs"] <= 91  # 76 +-20% of avg num_contigs
+    assert 4183 <= df.iloc[13]["sample_contigs"] <= 6273  # 5228 +-20% of avg largest_contig
+    assert 59515 <= df.iloc[14]["sample_contigs"] <= 89271  # 74393 +-20% of avg total_len
```
danejo3, line 191: https://github.com/bioforensics/yeat/issues/27#issuecomment-1301009696

This is my take on this suggestion.

I was pretty liberal on my +- buffer range.

The way I determined my medium for each assert is:
1. I ran the above list of arguments 5 times, 
2. took the average for `num_contigs`, `largest_contig`, and `total_len` and 
3. calculated the buffer +- caps.

Above the function, there is a decorator. When this function is executed with pytest, the function is called 3 times. Since, the seed is random by default, we do not need to specify the seed.

environment.yml
```diff
@@ -4,13 +4,14 @@ channels:
...
 dependencies:
-    - black=21.10b0
+    - black=22.10
     - fastp>=0.23
     - fastqc>=0.11
     - gzip>=1.7
     - mash>=2.3
     - megahit>=1.2
     - pytest-cov>=3.0
+    - python>=3.9
```
danejo3, line 14: YEAT cannot install if the user's python version is < 3.9. Add this to allow users to upgrade if needed.

danejo3:
Okay, code is ready for review! Let me know if you have any questions or concerns. Thanks

environment.yml
standage: Might want to add or update an entry in the change log describing why only Python >=3.9 is supported now.

yeat/cli.py
```diff
@@ -138,6 +140,13 @@ def get_parser(exit_on_error=True):
         default=150,
         help="target an average depth of coverage Cx when auto-downsampling; by default, C=150",
     )
+    parser.add_argument(
+        "--seed",
+        type=int,
+        metavar="S",
+        default=None,
+        help="override the randomly chosen seed S when downsampling; by default, S=None",
+    )
```
standage, line 149: The metavariable is useful when describing default numerical or string values, but in this case `S=None` doesn't convey the corresponding behavior. I'd suggest describing the default behavior instead. For your consideration:

> seed for the random number generator used for downsampling; by default the seed is chosen randomly

yeat/tests/test_cli.py
standage: Looks good! A couple comments.

I have no idea what information `df.iloc[12]["sample_contigs"]` stores. There *is* a describing its contents at the end of the line, but comments have a habit of coming out of sync with the code they are intended to describe. Probably better for legibility and clarity to assign those values to descriptive variable names before the assertion tests.

I think the `x <= var <= y` is a clear construction, but another you may consider uses `pytest.approx`. I use this most frequently to test the value of floating point numbers, for which simple `==` equality tests often fail (even if you're looking for an "exact" value, you have to specify some level of tolerance). But you can apply the same idea here, and just specify a wide tolerance. The first line would then become something like this, which is a pretty clear representation of 76 +/- 15.

```python
assert num_contigs == pytest.approx(76, abs=15)
```

standage (changes requested):
LGTM. See my comments below.

environment.yml
```diff
@@ -4,13 +4,14 @@ channels:
     - bioconda
     - defaults
 dependencies:
-    - black=21.10b0
+    - black=22.10
```
danejo3, line 7: Black version `21.10b0` has having incompatibilities errors with click. If users have click version `>8.1`, black will crash with:

> ImportError: cannot import name '_unicodefun' from 'click'

To fix this, users will need to downgrade click down to `8.0`.

This problem has been fixed in Black `22.3` and up.

danejo3:
Okay! A couple of comments on version pinning, added suggestions (Thanks!), and updated change log. Everything is ready to go!

environment.yml
standage: It doesn't much matter which version of Black is used, as long as it's used consistently. So you're welcome to upgrade and pin a newer version that doesn't have these issues. But that's often best left to a dedicated thread, since it can result in numerous trivial formatting changes that add a lot of noise and clutter to an existing PR.

standage (approved):

PR #30, opened 2022-11-03
Adding `--genome-size` flag to override auto-calculated genome size

### LINKED ISSUE #28: Add a flag to override genome size estimation
Sometimes when YEAT does the `MASH` `sketch` to determine the genome size estimation, the value can be wrong. If this is the case, users do not have the ability to override the estimation. A request has been made to create a flag to override genome size with their own input.

danejo3:
The purpose of this PR is to resolve #28 .

Sometimes, when YEAT auto calculates the genome size with MASH, the value can be off or if the user knows exactly the genome size, it's better to go with that number when appropriate.

In this PR, users can call the `--genome-size` flag to override YEAT's auto calculated size with the user's input.

### DIFF
--- a/yeat/cli.py
+++ b/yeat/cli.py
@@ -146,7 +146,7 @@ def get_parser(exit_on_error=True):
         type=int,
         metavar="G",
         default=0,
-        help="genome size; by default, G=0",
+        help="provide known genome size in base pairs (bp); by default, G=0",
     )
     parser.add_argument(
         "--init",
--- a/yeat/cli.py
+++ b/yeat/cli.py
@@ -46,6 +46,7 @@ def run(
     dryrun="dry",
     downsample=0,
     coverage=150,
+    genomesize=0,
 ):
     snakefile = resource_filename("yeat", "Snakefile")
     r1 = Path(fastq1).resolve()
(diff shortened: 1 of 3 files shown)


yeat/cli.py
```diff
@@ -138,6 +140,14 @@ def get_parser(exit_on_error=True):
...
         help="target an average depth of coverage Cx when auto-downsampling; by default, C=150",
     )
+    parser.add_argument(
+        "-g",
+        "--genome-size",
+        type=int,
+        metavar="G",
+        default=0,
+        help="genome size; by default, G=0",
+    )
```
standage, line 150: Genome size *in base pairs*, right? If so, I'd make sure to include the units in the help message. Sometimes programs expect or support users specifying genome sizes in Kbp, Mbp, and/or Gbp.

standage (changes requested):
Minor comment on the CLI documentation, otherwise this looks good!

danejo3:
Okay! This is a relatively simple PR and ready to ship!

standage (approved):

standage:
> This is a relatively simple PR

Indeed, but not so simple that we can skip running the tests. :smiley:

```
SyntaxError in line 90 of yeat/Snakefile:
invalid syntax
```

danejo3:
oops

PR #31, opened 2022-11-07
Accepting uncompressed fastq files

### LINKED ISSUE #26: Input is assumed to be gzipped - FastQC crashes if it is not
If non-gzipped files are given, they are copied to a location where they have the `.gz` extension. This confuses `fastqc` and it crashes.

Our solution was to gzip the input the files in place and re-run, but `yeat` probably shouldn't do that. My preferred solution would be to copy the files with the `.gz` extension only if they are gzipped to start. When the reads get to `fastp`, sending the filtered reads to an output with the `.gz` extension will trigger compression, so no rules using the filtered reads as input would have to change.

danejo3:
The purpose of this PR is to resolve #26 .

To run YEAT, the user is required to provide two _compressed_ fastq files (ie., `.gz`). 

Instead of forcing compressed input files, this PR aims to allow the flexibility of users to pass in both compressed and/or uncompressed reads.

To do this, a snakemake rule, `copyinput`, was updated. 

If users passed in an uncompressed file(s), YEAT will automatically compress it after the file has been copied over into the `seq/input` directory.

This early compression will prevent `fastp` from crashing because of the incompatible extensions assigned to the uncompressed file was passed in through the CLI.


yeat/Snakefile
```diff
@@ -27,9 +27,25 @@ rule copyinput:
...
+            copyfile(input.read2, output.read2)
+            shell("gzip {params.file1}")
+        elif input.read1[-3:] == ".gz" and input.read2[-3:] != ".gz":
+            copyfile(input.read1, output.read1)
+            copyfile(input.read2, params.file2)
+            shell("gzip {params.file2}")
+        else:
+            copyfile(input.read1, params.file1)
+            copyfile(input.read2, params.file2)
+            shell("gzip seq/input/*")
```
danejo3, line 48: The if statement is broken up into four different conditions. Couldn't think of an easier and cleaner way to satisfy the snakemake rule.

danejo3: Also, in this same area, the only way I am checking the compression for each input file is their extension. I could take a look into the reading the first few [bytes](https://stackoverflow.com/questions/6059302/how-to-check-if-a-file-is-gzip-compressed) of the file. Overkill?

danejo3 (commented):
Okay, first time doing a review of my own code! Let me know what you think! @standage

danejo3:
Adding @lovettse to this PR as well.

danejo3:
Hold on; I messed my code up. Sorry.

danejo3:
Okay! Ready! Sorry!

yeat/Snakefile
lovettse: I think trusting the extension is reasonable. This looks good to me.

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - 3.9 Python version pin for `exit_on_error` parameter and 3.10's incompatibility with SPAdes <v3.15.4 (#23, #29)
 - Custom genome size flag (#30)
 
+### Fixed
+- Uncompressed input files (#31)
```
standage, line 23: "Support for uncompressed input files"

yeat/Snakefile
```diff
@@ -28,8 +28,18 @@ rule copyinput:
...
         read2=config["read2"]
     run:
-        copyfile(input.read1, output.read1)
-        copyfile(input.read2, output.read2)
+        if input.read1[-3:] == ".gz" and input.read2[-3:] == ".gz":
+            copyfile(input.read1, output.read1)
+            copyfile(input.read2, output.read2)
+        elif input.read1[-3:] != ".gz" and input.read2[-3:] == ".gz":
+            shell("gzip -c {input.read1} > {output.read1}")
+            copyfile(input.read2, output.read2)
```
standage, line 36: A couple of comments.

1. You might be interested in the `str.endswith()` method, e.g. `filename.endswith(".gz")`
2. Is there any particular benefit or reason you're handling each possible combination of gzipped and non-gzipped input files, rather than handling each file on its own? If it works, something like the following would be a more clear and concise way to handle this.

```python
for inreads, outreads in zip(input, output):
    if inreads.endswith(".gz"):
        copyfile(inreads, outreads)
    else:
        shell(f"gzip -c {inreads} > {outreads}")
```

standage (changes requested):
Please see my comments. Also, we'll want to include a regression test as well.

yeat/Snakefile
danejo3: These are great suggestions! I didn't realize that we could `zip()` our `input` and `outputs` like that.

yeat/tests/test_cli.py
```diff
@@ -195,3 +196,32 @@ def test_random_downsample_seed(execution_number, capsys, tmp_path):
...
+
+@pytest.mark.long
+@pytest.mark.parametrize(
+    "inread1,inread2",
+    [
+        ("short_reads_1.fastq", "short_reads_2.fastq"),
+        ("short_reads_1.fastq.gz", "short_reads_2.fastq"),
+    ],
+)
+def test_uncompressed_input_reads(inread1, inread2, capfd, tmp_path):
```
danejo3, line 209: Added a test that tests:

1. 2 uncompressed files
2. 1 uncompressed + 1 compressed files

yeat/tests/test_cli.py
```diff
@@ -195,3 +196,32 @@ def test_random_downsample_seed(execution_number, capsys, tmp_path):
...
+        data_file(inread2),
+        "--outdir",
+        wd,
+    ]
+    args = cli.get_parser().parse_args(arglist)
+    cli.main(args)
+    outread1 = Path(wd).resolve() / "seq" / "input" / "sample_R1.fq.gz"
+    outread2 = Path(wd).resolve() / "seq" / "input" / "sample_R2.fq.gz"
+    assert outread1.exists()
+    assert outread2.exists()
```
danejo3, line 223: Checks if the copied/compressed files were named correctly and exists.

yeat/tests/test_cli.py
```diff
@@ -195,3 +196,32 @@ def test_random_downsample_seed(execution_number, capsys, tmp_path):
...
+    subprocess.run(["gzip", "-tv", outread1, outread2])
+    captured = capfd.readouterr()
+    assert re.search(r"seq\/input\/sample_R1.fq.gz:\s*OK", captured.err)
+    assert re.search(r"seq\/input\/sample_R2.fq.gz:\s*OK", captured.err)
```
danejo3, line 227: Run `gzip -tv` (-t, test if the compression was successful; -v, verbose).

danejo3 (commented):
Added tests, updated code with suggestions, all tests passing. Thanks for the suggestions!

yeat/Snakefile
danejo3: Updated code with Daniel's [suggestion](https://github.com/bioforensics/yeat/pull/31#discussion_r1015825043).

standage (commented):
This all looks good! Small PR incoming.

standage (approved):

PR #32, opened 2022-11-08
Eliminate the need for multiple copies of each file

standage:
Hi @danejo3, things in #31 look good. This PR includes a small suggested change. It eliminates the new uncompressed copies of the FASTQ files already present, which increases the required storage capacity for the repo by several fold. Instead, I've added a prep function for the test in question that will decompress each input file on the fly, if needed, to achieve the same result. For your consideration...


danejo3:
Oh good idea. I was concerned about the size increase, but this is a better way of handling the situation. Thanks!

PR #34, opened 2022-11-22
Adding Bandage to YEAT's workflow

### LINKED ISSUE #24: Adding Bandage to visual genome assembly
Bandage (`Bioinformatics Application for Navigating De novo Assembly Graphs Easily`) is a GUI for visualizing the quality of various assemblers when recreating the genome. It was created by Ryan Wick, who also created `Unicycler`, and is one of the go-to tools for many in the community.

In order to use Bandage, users will need to provide `assembly graph` files. These files are used to represent the final assembly of a genome. Typically, popular assemblers like SPAdes will auto-generate them alongside the final contigs. For example, users will find an `assembly_graph.fastg` file in their output directory.

The following assembly graph files are supported by Bandage:
- LastGraph (Velvet)
- FASTG (SPAdes)
- Trinity.fasta
- ASQG
- GFA

**BASIC USAGE:** 

To use the see assembly result, for iOS, users will click: 
> File -> Load graph -> {Locate the assembly graph file}
> After loading the graph, on the left hand side, click on the `Draw graph`--underneath the `Draw Graph` section.

image

(If you would like to learn more about what Bandage can do, look [here](https://rrwick.github.io/Bandage/), underneath the `Features` section.)

danejo3:
The purpose of this PR is to resolve #24--to add Bandage to the workflow. 

Bandage is a genome assembly viewer that allows you see the results of an assembly.

In this PR, another snakemake file, to run Bandage, was created. In this workflow, all fastg and gfa files from all assemblers will now create a jpeg file for each individual file.

In order for the user to enable Bandage, users must manually download or create the binary files to run it. Instructions to install Bandage is found in the the README. (Further instructions to install Bandage from the source code was deferred to Bandage's Q/A repo.)

Instead of creating an automatic way to install Bandage, a decision between @standage and @danejo3 was made to ask the users to download it themselves because the installation process for Bandage varies from one operating system to another.

If the user does not have Bandage in their $PATH, YEAT will skip this workflow and complete normally without it.


Makefile
```diff
@@ -27,6 +28,18 @@ style:
...
+	@if [[ "$(OS)" == "Linux" ]]; then \
+		yes A | unzip yeat/data/Bandage_Ubuntu_dynamic_v0_8_1.zip -d ~/Bandage; \
+		sudo apt install build-essential git libgl1-mesa-dev libxcb-xinerama0; \
+		sudo apt-get install libqt5gui5; \
+	elif [[ "$(OS)" == "Darwin" ]]; then \
+		yes A | unzip yeat/data/Bandage_Mac_v0_8_1.zip -d ~/Bandage; \
+	else \
+		echo "yeat does not support Bandage for: $(OS)"; \
+	fi
+
```
danejo3, line 42: > Downloaded the Bandage executables for both Linux and macOS and included them in the yeat repo.

I downloaded the Bandage executables for both Linux and macOS because of these various reasons:
1. I was unable to `conda install -c bioconda bandage`.
2. If you take a look at the last upload in anaconda, it says it was "2 years and 6 months ago".
3. When you try to conda install, you'll run into a problem where the package cannot be found in anaconda.
4. I think this stems from the problem that the author, Ryan, has said: "[He does not have] the time to give it the attention it deserves" and "don't expect a lot development or support".

To temporarily fix the problem, I added the Bandage executables from [here](https://rrwick.github.io/Bandage/), and created various commands to enable bandage in their home directory.

danejo3: As of right now, I plan on supporting both Linux and macOS. For both operating systems, if you run the entire `yeat` workflow, everything should work if the `make bandage` command was successful. 

For macOS, downloading, adding to the home directory, and using it immediately out of the box was very easy. The reason why is that the binary files needed for bandage come with the zipped `yeat/data/Bandage_Mac_v0_8_1.zip` file.
As a result, mac users should run into very little problems (hopefully).

The same cannot be said for Linux. There are a lot of challenges getting it to work on Linux.
1. the zipped file for Bandage does not contain binary files.
- As a result, you need to install various other files in order for it to work.
2. WSL2 has various other challenges that you need to install and get compared to a dedicated Linux OS
3. You need sudo. I was unable to get anything working, let alone, install necessary packages without it.

danejo3: I haven't tested thoroughly for a Linux-only system but for WSL2, I have quite a bit of experience.

In order for Bandage to run on WSL2, you'll need to do the following commands. Make sure you have sudo because you can't install anything without it.

After the Bandage executable is unzipped, you'll need to run this command at the very least:
`sudo apt-get install libqt5gui5`

This qt5 library is needed in order for bandage to run on the commandline.

In order for the qt5 GUI to open up and run, you'll need to install this:
-`conda install -c conda-forge qt`

If you are running into missing packages, try running this:
-`sudo apt install build-essential git libgl1-mesa-dev libxcb-xinerama0`
(This information was found in the installation documentation on Bandage.)

Because WSL2 is subsystem and not the main system, in order to view the GUI in the main operating system you will need to use `xserver`.

"Xserver  is an application that manages one or more graphics displays and one or more input devices (keyboard, mouse, etc.) connected to the computer.

It works as a server and can run on the local computer or on another computer on the network. Services can communicate with the X server to display graphical interfaces and receive input from the user." (from [here](https://askubuntu.com/questions/7881/what-is-the-x-server))

I'll have to write up a thing about this latter down the road. Let me know if you are interested and I'll write up a guide ASAP.

yeat/bandage/Snakefile
```diff
@@ -0,0 +1,82 @@
...
+        "analysis/spades/assembly_graph_after_simplification.gfa",
+        "analysis/spades/assembly_graph_with_scaffolds.gfa",
+        "analysis/spades/assembly_graph.fastg",
+        "analysis/spades/strain_graph.gfa"
+    params:
+        bandage=config["bandage"]
+    run:
+        for gfa in input:
+            filename = Path(gfa).stem
+            shell("{params.bandage} image {gfa} analysis/bandage/spades/{filename}.jpg")
```
danejo3, line 34: As noted in #24, we can use Bandage CLI to 
1) save bandage images, 
2) load a bandage plot on QT5, 
3) created a reduced subgraph of the input graph, and 
4) get a particular query and display that query only

In this PR, I took all of the `gfa` and `fastg` files found in all assemblers and converted them into `jpeg` files in `analysis/bandage/[assembler]`.

When using Bandage's CLI, there are only two main options used to display the graphs. Depending on what we want or what we are looking for, we can customize the bandage images by changing some visual effects.

The other two options of reducing the graph and viewing a specific query are interesting tools as well.

danejo3:
@standage @lovettse Let me know what you guys think. I'd recommend running this particular protype branch on a macOS for easy installation. I'll do more testing and debugging for Linux-based systems ASAP.

yeat/bandage/Snakefile
danejo3: I haven't played around with points 3 and 4, but I will get on it to show you guys some output examples.

yeat/bandage/bandage.py
```diff
@@ -0,0 +1,55 @@
...
+import platform
+from pkg_resources import resource_filename
+from snakemake import snakemake
+import warnings
+
+
+bandage = {
+    "Linux": Path("~/Bandage/Bandage").expanduser(),
+    "macOS": Path("~/Bandage/Bandage.app/Contents/MacOS/Bandage").expanduser(),
+}
```
standage, line 20: If a user wants to do a custom installation, they're welcome to install the program into a custom directory and set their `$PATH` accordingly. But by default we should assume that the `Bandage` executable is in the environment `$PATH`.

Makefile
```diff
@@ -27,6 +28,18 @@ style:
...
+bandage:
+	@if [[ "$(OS)" == "Linux" ]]; then \
+		yes A | unzip yeat/data/Bandage_Ubuntu_dynamic_v0_8_1.zip -d ~/Bandage; \
+		sudo apt install build-essential git libgl1-mesa-dev libxcb-xinerama0; \
+		sudo apt-get install libqt5gui5; \
+	elif [[ "$(OS)" == "Darwin" ]]; then \
+		yes A | unzip yeat/data/Bandage_Mac_v0_8_1.zip -d ~/Bandage; \
+	else \
+		echo "yeat does not support Bandage for: $(OS)"; \
+	fi
```
standage, line 41: I'd recommend unzipping into the conda environment rather than `~/Bandage/`. For macOS, you may even consider symlinking the Bandage executable to `$ENVDIR/bin/Bandage`.

yeat/bandage/bandage.py
```diff
@@ -0,0 +1,55 @@
...
+    if os not in compatable:
+        warnings.warn(f"yeat cannot support Bandage on this operating system: {os}")
+        return False
+    if os == "Linux" and not bandage["Linux"].exists():
+        warnings.warn("Bandage does not exist!")
+        return False
+    elif os == "macOS" and not bandage["macOS"].exists():
+        warnings.warn("Bandage does not exist!")
+        return False
+    return True
```
standage, line 35: If the executable is unzipped into the env directory (and copied/linked to the `$PATH` as needed), we could do a `$PATH` check here instead of looking at a fixed path in the user directory.

yeat/bandage/bandage.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def run_bandage(assembly_configs, outdir=".", cores=1):
+    os = platform.platform().strip().split("-")[0]
+    if not check_bandage_compatability(os):
+        return
```
standage, line 41: I don't think "compatibility" isn't the right term here: you're checking for availability AFAICT. But certainly I like this approach to handling a missing program gracefully.

Makefile
standage: Also, unless I'm missing something, the backslash and semi-colon characters are redundant. The backslash is to allow a single command to span multiple lines (it literally escapes the newline character), whereas the semicolon marks the end of one command and the beginning of another. Unless you have a good reason to, I'd recommend cleaning that up a bit.

standage: Also also, apt/apt-get is Ubuntu specific. That's likely going to be the majority of Linux users, but not all. We should give priority support to CentOS and RedHat. If that requires listing OS-level dependencies not easily installed by conda, then I suggest we move that to installation instructions rather than handling every OS flavor in the Makefile.

standage (changes requested):
Preliminary review below. I've successfully run yeat with Bandage plots, and the plots are great. Please see the feedback w.r.t. installation.

README.md
```diff
@@ -12,6 +12,24 @@ conda activate yeat
...
+export PATH={BANDAGE_DIR}:$PATH
+```
+
+Example for MacOS:
+```
+export PATH={BANDAGE_DIR}/Bandage.app/Contents/MacOS:$PATH
+```
+
+In order to run the pre-built binary files successfully, ensure that the binary file is kept in the same directory with all of the other files that came with it.
+
```
danejo3, line 32: Installation instructions for Bandage was added here.

Examples to update the user's $PATH is up for changes if needs be.

Not sure what else to say or grab's the user's attention. Hopefully the last line is informative because it took me a while to realize that each of the binary files requires various other files within the same dir.

yeat/bandage/bandage.py
```diff
@@ -0,0 +1,38 @@
...
+# -------------------------------------------------------------------------------------------------
+
+from pkg_resources import resource_filename
+from snakemake import snakemake
+import subprocess
+import warnings
+
+
+def check_bandage():
+    completed_process = subprocess.run(["Bandage", "--help"], capture_output=True, text=True)
```
danejo3, line 17: I am running ["Bandage", "--help"] instead of just ["Bandage"] for a reason.

When running just ["Bandage"], the command will open up the GUI and the GUI will put a pause/hold in the terminal. As a result, the snakemake workflow pauses until the user kills the GUI.

By doing ["Bandage", "--help"], we get the Bandage CLI and the help info.

Overall, I did this command because this command will tell YEAT and the user if Bandage is installed or not.

If it is not installed, the user will find this error message saying that "Bandage is not a directory!"

yeat/bandage/bandage.py
```diff
@@ -0,0 +1,38 @@
...
+    if completed_process.returncode == 1:
+        print(completed_process.stderr)
+        return False
+    return True
```
danejo3, line 21: When Bandage is not found in the users' $PATH, I have it print out the error message caught by subprocess.run().

I also when had this function return a boolean value based on the return code.

If the command was successful, the return code is 0--not 1. If failed, the return code is 1. (Makes sense.)

For the sake of readability when the function returns whether the command worked, instead of returning the return code, I returned back true or false accordingly.

yeat/bandage/bandage.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def run_bandage(assembly_configs, outdir=".", cores=1):
+    if not check_bandage():
+        warnings.warn("Unable to run Bandage; skipping Bandage")
```
danejo3, line 26: Warning message can be changed to anything. 

If command Bandage was unable to run successfully, we exit gracefully and finish YEAT's workflow.

yeat/tests/test_bandage.py
```diff
@@ -0,0 +1,35 @@
...
+# -------------------------------------------------------------------------------------------------
+
+import pytest
+import subprocess
+from unittest.mock import patch
+from yeat.bandage import bandage
+
+
+def test_checking_bandage():
+    assert bandage.check_bandage()
```
danejo3, line 17: This test is simple but I thought it was worth some merit to include it. If the developer does not have Bandage in  their path, this assert statement will fail.

Even if this test was excluded, all of tests that run the entire YEAT workflow will fail because the workflow includes the Bandage workflow, which requires the $PATH to be updated.

Maybe I should create a --bandage-off flag to disable it?

yeat/tests/test_bandage.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@patch("subprocess.run")
+def test_unable_to_run_bandage(function_mock, capsys):
+    completed_process = subprocess.CompletedProcess(["bandage", "--help"], 1)
+    error_message = "FileNotFoundError: [Errno 2] No such file or directory: 'Bandage'"
```
danejo3, line 23: This is error message displayed if the $PATH for Bandage is not set up.

danejo3 (commented):
Updates to the PR with previous suggestions, discussions, and my thoughts were made in the recent pushes. Let me know what you think! Thanks!

Makefile
danejo3: UPDATE: Users will need to download Bandage themselves, but will keep this thread open because some of the things you need to install for Ubuntu are relevant to get Bandage to work with the already made binary file.

yeat/bandage/bandage.py
danejo3: UPDATE: Dropped this idea. Making users update their $PATH accordingly to enable the Bandage command for their OS.

Makefile
danejo3: > For macOS, you may even consider symlinking the Bandage executable to $ENVDIR/bin/Bandage.

Tried to symlink; however, the binary file depends on all of the other files inside of the prepackaged directory. As a result, Bandage crashes.

Dropped the `make bandage` command. Making users update their $PATH accordingly to enable the Bandage command for their OS.

yeat/tests/test_bandage.py
standage: You're on the right track with this test, but I think you got a little too clever. You directly set the error message, so you're not *really* testing that this is the error message that gets produced in runtime, you're just checking that it's correctly propagated to the terminal (stdout). In other words, you could change the error message to "r0@$t b33f wh@ck@d00dl3" and the test would still pass: not an encouraging sign. :smiley:

I think the problem you're trying to solve here is how to test the "file not found" warning message when Bandage is installed and YEAT can in fact find it. If that's right, please revisit this test and see if you can find a way to elicit a genuine failure (manipulate the `$PATH` maybe?). If not, please let me know where I'm off.

yeat/tests/test_bandage.py
standage: I guess it depends on whether you want to make Bandage mandatory for running the test suite. That would be the easiest approach from the developer standpoint. But it would cause the test suite to fail for users that can't or don't want to install Bandage. And since YEAT will work fine without Bandage, you may want to accommodate that.

Consider using `pytest.mark` to mark this test. You could then run something like `pytest -m bandage` or `pytest -m 'not bandage'` like we do with "drmaa" for Mystt. (Of course, you'd probably want a Makefile shortcut for that.) This will facilitate running the full test suite when Bandage is installed, but excluding this test when it is not.

README.md
```diff
@@ -12,6 +12,24 @@ conda activate yeat
 pip install .
 ```
 
+### Bandage
+
+In order to run [`Bandage`](https://github.com/rrwick/Bandage) in YEAT's workflow, users will need to manually download the pre-built binaries files [here](https://rrwick.github.io/Bandage/) or build it from the source code [here](https://github.com/rrwick/Bandage). Instructions for building Bandage from the source code can be found [here](https://github.com/rrwick/Bandage#building-from-source).
+
+If the pre-built binary files are used, users will need to update their environment's `$PATH` to include the path of the directory containing the binary file.
```
standage, line 19: The `$PATH` needs to be set whether the user uses pre-built binaries or builds the binaries from scratch, right?

standage (changes requested):
A few more comments, questions, and suggestions for the code and documentation.

Makefile
```diff
@@ -9,15 +9,19 @@ help: Makefile
...
 
-## testall:     run all tests
+## testbandage: run only bandage required automated tests
+testbandage:
+	pytest --cov=yeat -m bandage
+
+## testall:     run all tests but bandage required tests
 testall:
-	pytest --cov=yeat
+	pytest --cov=yeat -m 'not bandage'
```
danejo3, line 24: When the user/developer runs `make testall`, no bandage required tests will be included in the run.

Instead, users will have to manually run bandage tests with `make testbandage`.

yeat/tests/test_bandage.py
```diff
@@ -0,0 +1,60 @@
...
+    subprocess.run(command, check=True)
+    assert jpg.exists()
+
+
+def test_convert_contig_to_fastg(tmp_path):
+    contig = data_file("k29.contigs.fa.gz")
+    fastg = tmp_path / "k29.contigs.fastg"
+    command = f"megahit_toolkit contig2fastg 29 {contig} > {fastg}"
+    subprocess.run(command, shell=True, check=True)
+    assert fastg.exists()
```
danejo3, line 60: Added a few more tests to 1) test `megahit`'s contig to fastg file converter and 2) test bandage's image converter.

danejo3 (commented):
Added a few more tests, updated wordings, and implemented suggestions. Thanks for the review!

README.md
danejo3: > If the pre-built binary files are used, users will need to update their environment's `$PATH` to include the path of the directory containing the binary file.

Updated this sentence to...

> Once the binary file has been obtained, users will need to update their environment's `$PATH` to include the path to the directory containing the binary file.

Misunderstood that when building from the source, it creates a binary file, which still requires the user to update their $PATH.

yeat/bandage/bandage.py
danejo3: `check_bandage_compatability()` function has been dropped.

yeat/tests/test_bandage.py
danejo3: A few tests now have the `bandage` pytest.mark decorator. All tests but `make testbandage` will exclude bandage required tests.

yeat/tests/test_bandage.py
danejo3: Thanks for the suggestion!

Found a solution to mock the $PATH variable.

yeat/bandage/Snakefile
```diff
@@ -0,0 +1,76 @@
...
+    run:
+        fastg_files = []
+        for fa in input:
+            kmer = re.findall(r'\d+', fa)[0]
+            filename = Path(fa).stem
+            fastg_files.append(f"analysis/megahit/intermediate_contigs/{filename}.fastg")
+            shell("megahit_toolkit contig2fastg {kmer} {fa} > analysis/megahit/intermediate_contigs/{filename}.fastg")
+        for fastg in fastg_files:
+            filename = Path(fastg).stem
+            shell("Bandage image {fastg} analysis/bandage/megahit/{filename}.jpg")
```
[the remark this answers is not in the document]
danejo3: To be honest, I am unsure why I did this? The only thing that I could think of was the uniformity of the terminal's output.

First we print out that all the contig files are being converted and then we print out the image creation for each fastg file.

With the single loop, we changed the display to:

convert contig
create jpg
convert contig
create jpg
etc. etc. etc.

Previously...

convert contig
convert contig
etc. etc. etc.

create jpg
create jpg
etc. etc. etc.

yeat/bandage/Snakefile
```diff
@@ -0,0 +1,76 @@
...
+        shell("touch {output[0]}")
+
+
+def get_unicycler_graphic_files(wildcards):
+    glob_pattern = os.path.join('analysis/unicycler', '*.gfa')
+    files = sorted(glob(glob_pattern), key=os.path.getctime)
+    return files
```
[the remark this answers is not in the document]
danejo3: This is so much better!

standage (approved):
Code looks much cleaner. I will perform a functional test tomorrow and let you know!

standage:
I just ran SPAdes in `--isolate` mode and the Bandage step choked and failed because it couldn't find `analysis/spades/strain_graph.gfa`. You might want to use globbing to test which graph files are present, rather than hardcoding a list of files that may change depending on modes or parameters.

```
MissingInputException in rule spades in line 21 of /Users/daniel.standage/Software/yeat/yeat/bandage/Snakefile:
Missing input files for rule spades:
    output: analysis/bandage/spades/.done
    affected files:
        analysis/spades/strain_graph.gfa
Traceback (most recent call last):
  File "/Users/daniel.standage/anaconda3/envs/yeat/bin/yeat", line 33, in <module>
    sys.exit(load_entry_point('yeat', 'console_scripts', 'yeat')())
  File "/Users/daniel.standage/Software/yeat/yeat/cli.py", line 186, in main
    bandage.run_bandage(
  File "/Users/daniel.standage/Software/yeat/yeat/bandage/bandage.py", line 42, in run_bandage
    raise RuntimeError("Snakemake Failed")
RuntimeError: Snakemake Failed
```

yeat/bandage/Snakefile
```diff
@@ -0,0 +1,64 @@
...
+    pattern = r"analysis/megahit/intermediate_contigs/k\d+.contigs.fa"
+    contigs = glob(r"analysis/megahit/intermediate_contigs/k*.contigs.fa")
+    return filter(re.compile(pattern).match, contigs)
+
+
+rule megahit:
+    output:
+        "analysis/bandage/megahit/.done"
+    input:
+        get_and_filter_contig_files()
```
danejo3, line 43: As pointed out in the recent bug, to avoid hard-coded input file names, we are doing a glob search now in the rule `megahit` as well.

When applying this change, we have to ensure that the contig files that `megahit_toolkit` converts are the `k*.contigs.fa`--nothing more than that! In `analysis/megahit/intermediate_contigs`, megahit produces a lot of different`contigs.fa` files for each k-mer. For example, `k[blah].final.contigs.fa`. These "final" files cannot be converted into fastg files.

In consequence, I created a helper function called `get_and_filter_contigs_files()`.

In this function, I have a regex pattern and a glob function that retrieves all files matching with the Unix wildcard. I kept the glob retrieve function because it keeps the relative path of the return list of files and was the easiest to work with when doing the various shell commands.

Fun fact, glob only works with Unix wildcard input strings and not regex. In order to filter the list of glob'ed files, you need to do a regex filter.

I couldn't find another better way than this.

danejo3 (commented):
Okay! The patch for the problem that you ran into was fixed! Let me know if you run into anything else. Thanks! @standage

yeat/bandage/Snakefile
standage: Yeah, I'm not sure we'll get it much cleaner than that. Thanks!

Issue #35, opened 2023-01-13
Adding support for long read assembly algorithms


danejo3:
There are two common ways to sequence a DNA: short-read and long-read sequencing. As of right now, YEAT only supports short-read sequencing. Adding long-read assembly support will be a tremendous addition to YEAT's capabilities because long-read sequencing has come a long way, is rising in popularity, and can solve many of the downsides that short-read sequencing has.

While both techniques have their pros and cons, adding ability to assembly long reads would be a great way to diversify YEAT's capabilities.

To explore this long-read assembly algorithms, a few algorithms of interest came into my radar:
-[Flye](https://github.com/fenderglass/Flye) (all of it's variations: metaFlye and virally in particular)
-[Trycycler](https://github.com/rrwick/Trycycler)
-[Canu](https://github.com/marbl/canu)
-[Raven](https://github.com/lbcb-sci/raven)
-[Redbean](https://github.com/ruanjue/wtdbg2)

If we were to begin support for long-read assembly, I would start with Flye, for doing everything, and Trycycler, for their focus on bacterial genomes.

Flye is interesting because it can practically do anything with long reads. Similar to SPAdes!
> "It is designed for a wide range of datasets, from small bacterial projects to large mammalian-scale assemblies."

Flye also has a bunch of flags, like SPAdes, that adjusts the algorithm to produce better assembly results for metagenomics and plasmids.

Trycycler would be great to algorithm to add for its speciality in bacterial genome assembly. Not quite sure how it compares with Flye's `--plasmid` since the last publication back in 2021.

danejo3:
While Trycycler has not been implemented, will plan to add in the future, `Flye`, `Canu`, `Hifiasm` (#49), and `Hifiasm-meta` (#49) has been added.

Will look into `Raven` and `Redbean` more. It seems like `Raven` is actively developed while `Redbean` last release was 2019.

standage:
Yeah, we evaluated wtbd...wbtgb...wt...ugh, *Readbean* a few years ago. I don't remember if we ever did a thorough assessment, but the first-pass results weren't super promising IIRC. Since it doesn't appear to be actively developed, I wouldn't make this assembler a high priority for inclusion.

danejo3:
YEAT has grown a lot over the past year. Long read assemblers such as flye, canu, unicycler, hifiasm, hifiasm_meta, and metaMDBG were added.

If there are any new long read assembly algorithms to evaluate and add, let's start creating new issue threads.

PR #37, opened 2023-03-02
Adding support to do basic pacbio hifi-read assembly with Flye and Canu

danejo3:
The purpose of this PR is to begin tackling #35 .

In this PR, the ability to do basic assembly of PacBio HiFi-reads was introduced. To assemble these reads, two new algorithms were added: 1) Canu and 2) Flye.

To run these algorithms, YEAT requires an input flag stating what kind of reads are to be assembled.

For example,

> yeat --pacbio [hifi_data]

or when assembling paired-end short read data,

> yeat --paired [short_read1] [short_read2]

***When running Canu, users will need add `genomeSize=N` in the `extra_args` "Canu" section of the config. This parameter is in order to run the assembler.***

In this PR, YEAT still relies on the config file to fine-tune the assembler's algorithm.

For example, in the config, users can adjust the `correctedErrorRate` and other additional flags for Canu.

```
[
    {
        "algorithm": "canu",
        "extra_args": "correctedErrorRate=0.075 -trimmed -corrected"
    }
]
```

The major backbone of this PR is the snakemake file: `Pacbio.smk`. In this PR, we created a separate snakemake file from the paired-end reads assembly (`Paired.smk`) because of the differences in the number of required input and output expected files for each rule. In the future, creating a class to return back the inputs and outputs of each rule will help consolidate these snakemake file. 

Overall, the goal of this PR is to layout the basic workflow for pacbio hifi-read assembly.


danejo3:
Alrighty! I've got a pretty good PR ready for a prelim-review. @standage 

Here are some example usages:

Example for short read assembly
`yeat -o sandbox1 short --paired short1.gz short2.gz config.cfg`

Example for long read assembly (using Flye)
`yeat -o sandbox2 long --pacbio long.gz config.cfg`

Notice the specific ordering that the flags and inputs must be in.

`yeat [options] {subcommand} {type of data} {read(s)} {config file}`

To view the help info for each subcommand:

`yeat {short | long} -h`

As of right now, YEAT is only supporting long-read **fastq** files. I've learned that PacBio instruments produce 3 different kinds of files: 1) .bam, 2) .fasta, and 3) .fastq. 

Feel free to explore the new CLI.

I have enforced people to specify the type of reads that they are assembling and what kind it is as well.

Something that I would like to explore more are following avenues in the future.

Short-read assembly: Paired-end, Single-end, and Interleaved-end pairs
Long-read assembly: PacBio, Nanopor, and Sanger
Hybrid assembly

yeat/bandage/Snakefile
```diff
@@ -62,3 +62,29 @@ rule unicycler:
...
+            filename = Path(gfa).stem
+            shell("Bandage image {gfa} analysis/bandage/flye/{filename}.jpg")
+        shell("touch {output[0]}")
+
+
+rule canu:
+    output:
+        "analysis/bandage/canu/.done"
+    input:
+        glob("analysis/canu/unitigging/4-unitigger/*.gfa")
```
danejo3, line 85: Canu deprecated a lot of their "gfa" files since v2.1. The only gfa file they have left is "best edges" without repeating edges graph.

This is what skoren said:

> Unitigs + their gfa intentionally removed.

> They were just too fragmented (more fragmented than they needed to be) and unreliable (missing some edges) to be of useful. We expect in the future to add more reliable assembly graphs once we have finalized an improved contig strategy.

https://github.com/marbl/canu/issues/1786
https://github.com/marbl/canu/issues/1819

yeat/cli/cli.py
```diff
@@ -0,0 +1,108 @@
...
+    read_input_options = parser.add_argument_group("input read options")
+    mx = read_input_options.add_mutually_exclusive_group(required=True)
+    mx.add_argument(
+        "--paired",
+        metavar=("READ1", "READ2"),
+        type=str,
+        nargs=2,
+        help="paired-end reads in FASTQ format",
+    )
+    mx.add_argument("--pacbio", metavar="READ", type=str, help="PacBio HiFi-reads in FASTQ format")
```
danejo3, line 88: Created a mutually exclusive group where the user must state the type of read they are working with.

As of right now, YEAT requires the user to provide either: 1) --paired (for paired short-end reads) or 2) --pacbio (for PacBio HiFi-reads).

I decided to go this route instead of the previous implementation of subcommands because I would like to added additional read types in the future such as: 1) single-end reads (short-read), 2) interleaved-end pairs (short-read), and 3) Nanopore reads (long-reads).

Ideally, I would like this in the future:

| read-type command | input data |
| ------------- | ------------- |
| --paired  | [read1] [read2] |
| --single  | [read]  |
| --interleaved  | [read]  |
| --pacbio  | [read]  |
| --nanopore  | [read]  |

yeat/tests/test_pacbio.py
```diff
@@ -0,0 +1,71 @@
...
+        "--threads",
+        cores,
+        "--pacbio",
+        data_file("ecoli.fastq.gz"),
+        data_file(f"{algorithm}.cfg"),
+    ]
+    args = cli.get_parser().parse_args(arglist)
+    cli.main(args)
+    result = Path(wd).resolve() / "analysis" / algorithm / finalcontigs
+    assert result.exists()
```
danejo3, line 71: This test takes way too long: ~30-40 mins.

I added the "pytest.mark" `extralong` to mark test that a super long.

yeat/workflows/workflows.py
```diff
@@ -0,0 +1,105 @@
...
+        raise RuntimeError("Snakemake Failed")
+
+
+def check_canu_required_params(extra_args, cores):
+    if "genomeSize=" not in extra_args:
+        raise ValueError("Missing required input argument from config: 'genomeSize'")
+    if cores < 4:
+        raise ValueError(
+            "Canu requires at least 4 avaliable cores; increase `--threads` to 4 or more"
+        )
```
danejo3, line 62: In order to run Canu, users must input the `genomeSize` of the sample.

| CLI parameter  | Input | Does YEAT provide automatically?
| ------------- | ------------- | ------------- |
| genomeSize=  | An estimate of the size of the genome. Common suffices are allowed, for example, 3.7m or 2.8g.  | No, users will need to provide this data in the `extra_params` "Canu" section in the config |

Users cannot use YEAT's `-g` or `--genome-size` because YEAT cannot handle shortcut suffices of genomes sizes--only the int allowed.

Furthermore, Canu requires **at least 4** available cores to run. If the computer, job, etc. does not have enough, Canu will fail to run.

Added a check statement making sure that the user puts at least 4 cores when filling out YEAT's CLI with `-t` or `--threads`.

danejo3 (commented):
Added a bunch of comments on the prelim review for this MR. 

Biggest concerns that I have is how to test long-read data and if the "mutually exclusive group" required parameter is the way to go for differentiating the input data types.

danejo3:
Long read ecoli test data was grabbed from [here](http://gembox.cbcb.umd.edu/mhap/raw/ecoli_p6_25x.filtered.fastq):

> curl -L -o pacbio.fastq http://gembox.cbcb.umd.edu/mhap/raw/ecoli_p6_25x.filtered.fastq

This dataset was used in the Canu tutorial.

The size of the file is ~2-3 GB. When gzipped, it drops down to 1.32 GB.

Not ideal for adding this test data to the package.

Makefile
```diff
@@ -9,20 +9,29 @@ help: Makefile
...
+	pytest --cov=yeat -m 'long and not bandage and not hifi'
 
 ## testbandage: run only bandage required automated tests
 testbandage:
 	pytest --cov=yeat -m bandage
 
+## testhifi:    run only PacBio HiFi-reads automated tests
+testhifi:
+	pytest --cov=yeat -m 'hifi and not long and not bandage'
+
```
danejo3, line 25: New make command: `testhifi`. This command will only run assemblers that use the pacbio hifi-reads. These tests are extremely long and can take up to an hour depending on the computers resources.

danejo3 (commented):
@standage This PR is ready for review! The `make testhifi` command will take a long time to execute. Feel free to manually test yeat's pacbio hifi-read assembly manually by using the command:

> yeat -o sandbox -t 16 --pacbio {fastq_file} {config_file}

When running Canu, make sure you put the estimated genome size in the `extra_args` section of the config.

yeat/workflows/workflows.py
standage: Out of curiosity: does the Mash-based genome size estimation not work for hifi data?

standage (commented):
No concerns with the code from my perspective. Test is running on my machine. If there are no issues, I'll merge when it passes!

yeat/workflows/workflows.py
danejo3: I haven't tried yet! I completely forgot about that. Sean and I discussed about it and we both think it should work! I'll do some testing.

yeat/bandage/Snakefile
```diff
@@ -62,3 +62,29 @@ rule unicycler:
...
+        "analysis/bandage/flye/.done"
+    input:
+        glob("analysis/flye/30-contigger/*.gfa"),
+        glob("analysis/flye/40-polishing/*.gfa"),
+        glob("analysis/flye/*.gfa")
+    run:
+        for gfa in input:
+            filename = Path(gfa).stem
+            shell("Bandage image {gfa} analysis/bandage/flye/{filename}.jpg")
+        shell("touch {output[0]}")
```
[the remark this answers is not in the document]
danejo3: Oh, I apparently miscommented on the wrong code! Yes, this code is for Bandage!

yeat/bandage/Snakefile
```diff
@@ -62,3 +62,29 @@ rule unicycler:
...
+rule canu:
+    output:
+        "analysis/bandage/canu/.done"
+    input:
+        glob("analysis/canu/unitigging/4-unitigger/*.gfa")
+    run:
+        for gfa in input:
+            filename = Path(gfa).stem
+            shell("Bandage image {gfa} analysis/bandage/canu/{filename}.jpg")
+        shell("touch {output[0]}")
```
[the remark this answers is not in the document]
danejo3: Oh, I apparently miscommented on the wrong code! Yes, this code is for Bandage!

yeat/workflows/snakefiles/Pacbio
```diff
@@ -0,0 +1,82 @@
...
+    input:
+        read="seq/input/{sample}.fq.gz"
+    threads: 128
+    params:
+        extra_args=config["extra_args"]
+    shell:
+        """
+        flye --pacbio-hifi {input.read} -t {threads} -o analysis/flye {params.extra_args[flye]}
+        ln -s assembly.fasta {output.contigs}
+        """
```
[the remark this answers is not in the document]
danejo3: Oops. Yes, thanks for notice! Updated the environment.yml file.

yeat/workflows/workflows.py
danejo3: UPDATE: Did some testing with Mash's genome size estimation. Calculated genome sizes way over the actual genome size. Tried to adjust `mash sketch` parameters but was unable to get to the actual size.

The dataset that I used was PacBio's HiFi reads for [Ecoli K12](https://trace.ncbi.nlm.nih.gov/Traces/index.html?view=run_browser&acc=SRR10971019&display=metadata). Ecoli K12 has a genome size of ~4.6 mb (see paper [here](https://pubmed.ncbi.nlm.nih.gov/9278503/)). It is a very well studied bacteria. 

Theoretically, we should be able to calculate the genome size with PacBio reads and with high confidence. To calculate the genome size, we take the total number of k-mers / coverage. PacBio reads have a 99.9999% error-free rate. Because of this, we would expect a high accuracy in the total number of k-mers.

When running `mash sketch`, I got back a variety of over-estimated genome sizes. I adjusted the parameters to get to the expected size but was unable to.

Here are the results of my mash sketch commands:

| command  | parameters | estimated genome size | coverage |
| ------------- | ------------- | ------------- | ------------- |
| mash sketch [fastq]  | -r  | 33.6 mb | 40x |
| mash sketch [fastq]  | -k 32 -r  | 51.3 mb | 24x |
| mash sketch [fastq]  | -k 32 -m  3 -r  | 11 mb | 117x |
| mash sketch [fastq]  | -k 21 -m 3 -r  | 8.7 mb | 144x |

-r = read set
-k = k-mer size
-m = Minimum copies of each k-mer required to pass noise filter for reads.

Not sure why I am unable to get close to the expected genome size. Consulted with @lovettse, but was unable to figure it out.

---

To do:
1. try another dataset of pacbio hifi read data with known genome size to check the genome size estimation. 
2. test Illumina short read data and check if I can get the expected genome size estimation (did not test this in the past)
3. how much of an impact will a genome size with x2 or x3 the actual size will affect Canu's assembly? (Canu requires genome size.)
4. sourmash - a variant of MASH?

@lovettse mentioned something interest that I didn't think about before. Most bacteria genomes is around 7 kb. Maybe if the input data is bacteria we can set that value as default? (Very duck-tape answer.)

danejo3: Created issue thread #42 . I will revisit this problem in another PR. For this PR, let's keep it the way it is. I'll delve into estimating genome size problem in the future and integrate it in the workflow.

Issue #41, opened 2023-05-03
Enabling support for handling multiple samples


danejo3:
Currently, YEAT can only process one sample at a time. If a user has multiple samples that they want to process in a single run, adding a structured data directory or list of of sample names/paths would help enable this feature.

For example,

User wants to pass in SAMPLE1 and SAMPLE2 into the command line or...

Directory
->SAMPLE1
->SAMPLE2

With the directory, users can pass this in as well.

PR #43, opened 2023-05-09
New config file format; enables multiple input samples and running multiple assembly algorithms (both paired and pacbio!)

### LINKED ISSUE #40: Enabling support for running the same assemblers multiple times
One of the current design problems that YEAT has is its inability to execute the same assembly algorithm multiple times in a single run.

For example, if I wanted to run `spades` twice. One without the `--meta` flag and one with the `--meta` flag, Snakemake will see that the files and directories for the first run was created and will therefore not run the second `spades` job.

To fix this problem, in the config file, we can introduce a `label` key for each assembly algorithm that the user provides. Each label will need to be unique.

danejo3:
In this PR, a major overhaul of CLI was made because of the new proposed changes to the required config file. In the new config file, users will now input samples with file paths and assembly algorithms with labels, extra_args, and names of samples.

For example:
```
{
    "samples": {
        "sample1": [
            "/Users/dane.jo/Desktop/test1/data/short1.fq.gz",
            "/Users/dane.jo/Desktop/test1/data/short2.fq.gz"
        ],
	"sample2": [
            "/Users/dane.jo/Desktop/test1/data/short1.fq.gz",
            "/Users/dane.jo/Desktop/test1/data/short2.fq.gz"
        ],
	"sample3": [
	    "/Users/dane.jo/Desktop/test1/data/ecoli.fastq.gz"
	]
    },
    "assemblers": [
        {
            "label": "spades-meta",
            "algorithm": "spades",
            "extra_args": "--meta",
            "samples": [
                "sample1"
            ]
        },
        {
            "label": "megahit-mins",
            "algorithm": "megahit",
            "extra_args": "--min-count 5 --min-contig-len 300",
            "samples": [
                "sample1",
		"sample2"
            ]
        },
	{
            "label": "hicanu",
            "algorithm": "canu",
            "extra_args": "genomeSize=4.6m",
            "samples": [
                "sample3"
            ]
        },
        {
            "label": "hiflye",
            "algorithm": "flye",
            "extra_args": "",
            "samples": [
                "sample3"
            ]
        }
    ]
}
```

By using this new config file, this PR will attempt to resolve #40 and #41.

Because of the multi-input samples, the analysis directory has changed to organize all of the assembly results by sample names. Below is a screenshot of the new output directories.

Notice that Bandage and Quast are now in the final assembly directory. Also, because of the new config file format, users can now run both paired and long read assemblies!

image


danejo3:
@standage Whiteboard sesh

Before talking to Karla, for the past few days, I spent some time understanding how multi-input sampling is done in one of our internal workflows and tried to re-implement it for YEAT in `samples.py`. You can find this in the first [push](fc103f89825781fababaadb6f97c79063e7fbb95) of this PR.

The more I worked with it and tried to conform YEAT to it, the more I realized how incredibly difficult and self-imposing it was to force the same strategy onto YEAT.

Yesterday, I took a step back and asked myself, "If YEAT were independent of the current way we handle multi-input sampling, how would I go about it?"

In our discussion, we came up with this solution. This solution can resolve #40 and #41.

One of the neat features that YEAT has is its ability to assemble, for example, with multiple unique assembly algorithms. In #40, a feature request for enabling multiple assemblies on the sample algorithm was made. This is because, sometimes, if the sample is unknown, it can be helpful to run the same assembler multiple times with different flags or commands.

For example:
- spades
- spades-meta
- spades-krona
- etc...

To enable this feature, in the config file, an additional parameter would be added to each assembly algorithm called "label".

```
[
        {
		"label": "spades_normal",
		"algorithm": "spades",
		"extra_args": ""
	},
	{
		"label": "spades_meta",
		"algorithm": "spades",
		"extra_args": "--meta"
	},
	{
		"label": "spades_krona",
		"algorithm": "spades",
		"extra_args": "--krona"
	}
]
```

Along with these changes, issue thread #41 talks about enabling support for multiple input samples. This feature would be extremely helpful when you want to run YEAT with multiple samples in one run. (We'll need to enable grid support to divvy up the job.)

For example,
- `sample1` has 2 fastq files
- `sample2` has 2 fastq files
- `sample3` has 1 fastq file

For `sample1`, we want to runt he label "spades-meta". For `sample1` and `sample2`, we want to run the labels "megahit-mins". For `sample3`, we want to run the label `Pacbio-hifi`.

In the final proposed config file, the following would be created.

```
{
    "samples": {
        "sample1": [
            "path_to_read1",
            "path_to_read2"
        ],
        "sample2": [
            "path_to_read1",
            "path_to_read2"
        ],
        "sample3": [
            "path_to_read"
        ]
    },
    "assemblers": [
        {
            "label": "spades-meta",
            "algorithm": "spades",
            "extra_args": "--meta",
            "samples": [
                "sample1"
            ]
        },
        {
            "label": "megahit-mins",
            "algorithm": "megahit",
            "extra_args": "--min-count 5 --min-contig-len 300",
            "samples": [
                "sample1",
                "sample2"
            ]
        },
        {
            "label": "pacbio-hifi",
            "algorithm": "canu",
            "extra_args": "genomeSize=4.8m",
            "samples": [
                "sample3"
            ]
        }
    ]
}
```  

The reason why the config is laid out in this proposed manner is because:
1) there is less opportunities for users to introduce input error (tried to avoid users from copy pasting)
2) samples can be passed in from anywhere by providing a PATH
3) the CLI become immensely cleaner (for example, to run basic YEAT command: `yeat -o sandbox -t 16 config.cfg`)

Because of the config layout, the following changes to the snakemake file output directory structures will be made. Taking from the example above.

```
- seq
- -> blah
- analysis
  - sample1
     - spades-meta
     - megahit-min
  - sample2
     - megahit-min
  - sample3
     - Pacbio-hifi
```

danejo3:
Screenshot 2023-05-10 at 3 47 40 PM

standage:
The proposed config file format looks good to me, as does the corresponding working directory format. Just one comment based on your example: you might consider making QUAST a subdirectory of each sample, rather than having a single QUAST directory containing every sample. That seems more consistent with the proposed layout.

<small>Also, I think the term you're looking for is `spades.py --corona`, not `spades.py --krona`.</small>

danejo3:
Running into a bug with MEGAHIT. Not sure what the problem is.

When running MEGAHIT with `-t [2 or more cores]` on my machine, it seg faults with error code `11`.

Computer Specs:

image

In MEGAHIT's documentation:

image

Running MEGAHIT with `-t 2` failed:

image

I've tried running MEGAHIT with different varying sizes 3, 4, and 8, but same error. I've ran MEGAHIT by itself without the YEAT workflow but still getting the same error.

Interestingly, I've ran MEGAHIT on the server were we have plenty of cores and was able to run it successfully with `-t 128`.

yeat/tests/test_paired.py
```diff
@@ -7,83 +7,108 @@
...
+#     cli.main(args)
+#     analysis_dir = Path(wd).resolve() / "analysis"
+#     expected = [
+#         analysis_dir / "sample1" / "default_spades" / "spades" / "contigs.fasta",
+#         analysis_dir / "sample2" / "default_spades" / "spades" / "contigs.fasta",
+#         analysis_dir / "sample2" / "default_megahit" / "megahit" / "final.contigs.fa",
+#         analysis_dir / "sample1" / "Shigella_sonnei_plasmids" / "unicycler" / "assembly.fasta",
+#     ]
+#     for observed in expected:
+#         assert observed.exists()
```
danejo3, line 111: This test is extremely long--takes almost an hour. It runs 4 assemblers with two different data sets. The quickest one is `short_reads_*.fastq.gz` and the longest one is `Animal_289_R*.fq.gz`. I'm going to remove this test and the short read dataset of `Animal_289_R*.fq.gz` because they take too long to assemble for testing purposes.

yeat/cli/cli.py
```diff
@@ -57,13 +68,6 @@ def options(parser):
         default=".",
         help="output directory; default is current working directory",
     )
-    parser.add_argument(
-        "--sample",
-        type=str,
-        metavar="S",
-        default="sample",
-        help="specify a unique sample name S for storing assembly results in the working directory; by default S=sample",
-    )
```
danejo3, line 66: Removed this flag. We will now use the `labels` of each algorithm instead.

For example, "canu" requires the user to provide `-p`, the assembly prefix. Instead of using the original user's input `--sample`, we can use the algorithm's `label` in the config file.

yeat/cli/cli.py
```diff
@@ -72,37 +76,24 @@ def options(parser):
...
-        type=str,
-        nargs=2,
-        help="paired-end reads in FASTQ format",
+    parser.add_argument(
+        "-n",
+        "--dry-run",
+        action="store_true",
+        help="construct workflow DAG and print a summary but do not execute",
     )
-    mx.add_argument("--pacbio", metavar="READ", type=str, help="PacBio HiFi-reads in FASTQ format")
```
danejo3, line 88: No more need for this function with the new config file. In the config file, I expect users to provide the correct number of reads/files for each assembly algorithm. 

In future PR, I would like to do some input checking, ensuring that there are enough files, but for now, I would like to "trust" that the user knows what they are doing.

yeat/cli/cli.py
```diff
@@ -72,37 +76,24 @@ def options(parser):
...
 
 def main(args=None):
     if args is None:
-        args = get_parser().parse_args()
-    assembly_configs = AssemblerConfig.parse_json(open(args.config))
-    workflows.run_workflow(args, assembly_configs)
-    if not args.dry_run:
-        bandage.run_bandage(
-            assembly_configs=assembly_configs, outdir=args.outdir, cores=args.threads
-        )
```
danejo3, line 108: Moved the entire "bandage" workflow into the `workflows` directory.

`bandage.run_bandage()` will be called in the `run_workflows()` function.

yeat/cli/config.py
```diff
@@ -54,3 +50,36 @@ def validate(config):
...
+        self.label = label
+        if algorithm not in ALGORITHMS:
+            raise AssemblyConfigurationError(f"Unsupported assembly algorithm '{algorithm}'")
+        self.algorithm = algorithm
+        self.samples = samples
+        self.extra_args = extra_args
+
+    @classmethod
+    def from_json(cls, data):
+        return cls(data["label"], data["algorithm"], data["samples"], data["extra_args"])
```
danejo3, line 85: Went with an OOP route. Not sure if it made the code better or worse. The `Sample` and `Assembler` class objects are to represent portions or sections of the config. Within each object, error checking is done to ensure that the input config file has no user errors.

Example of how the OOP classes are used:
```
{
    "samples": {
        "sample1": SAMPLE_OBJ,
        "sample2": SAMPLE_OBJ,
        "sample3": SAMPLE_OBJ
    },
    "assemblers": [
        ASSEMBLER_OBJ,
        ASSEMBLER_OBJ,
        ASSEMBLER_OBJ,
        ASSEMBLER_OBJ,
        ASSEMBLER_OBJ
    ]
}
```

yeat/tests/data/configs/canu.cfg
```diff
@@ -0,0 +1,17 @@
...
+        {
+            "label": "canu-default",
+            "algorithm": "canu",
+            "extra_args": "genomeSize=4.8m",
+            "samples": [
+                "sample1"
+            ]
+        }
+    ]
+}
```
danejo3, line 17: Created a lot of config test files. As a result, to organize them, I created a dedicated `configs` directory in `yeat/tests/data`.

yeat/tests/data/configs/paired.cfg
```diff
@@ -0,0 +1,39 @@
...
+        "sample1": [
+            "yeat/tests/data/short_reads_1.fastq.gz",
+            "yeat/tests/data/short_reads_2.fastq.gz"
+        ],
+        "sample2": [
+            "yeat/tests/data/Animal_289_R1.fq.gz",
+            "yeat/tests/data/Animal_289_R2.fq.gz"
+        ]
+    },
+    "assemblers": [
```
danejo3, line 12: Example of a complex YEAT job assembling 4 times for 2 samples.

-2 jobs for spades
-1 job for megahit
-1 job for Unicycler

danejo3 (commented):
In this PR, I've added several tests to improve the overall coverage for YEAT. I have it at 97%.

danejo3:
Still running into this issue with running custom thread numbers for `MEGAHIT` on my MacBook Pro machine. Not sure why I'm segfault-ing. 

Need to support grid ASAP. Enabling multiple assembly runs is too time consuming when running on a single machine. 

This MR is ready for review! @standage 

Let me know if there are any changes you would like me to pursue or add. I think this PR will be a huge leap stone capability for YEAT once merged. Will probably cut a release after this or when grid support is enabled.

yeat/cli/config.py
```diff
@@ -8,44 +8,40 @@
...
 import json
+from pathlib import Path
 from warnings import warn
 
 
-ALGORITHMS = ("spades", "megahit", "unicycler", "flye", "canu")
-KEYS = ["algorithm", "extra_args"]
+ALGORITHMS = ["spades", "megahit", "unicycler", "flye", "canu"]
+KEYS1 = ["samples", "assemblers"]
+KEYS2 = ["label", "algorithm", "extra_args", "samples"]
```
standage, line 17: Small suggestion: if you define these constants as tuples instead of lists (as you did before with `ALGORITHMS`) you are making them immutable (as constants are expected to be) and protecting the data from both intentional and unintentional modification.

yeat/cli/config.py
```diff
@@ -8,44 +8,40 @@
...
+        assemblerlist = [Assembler.from_json(assembler) for assembler in configdata["assemblers"]]
+        return sampledict, assemblerlist
 
     @staticmethod
-    def validate(config):
-        missingkeys = set(KEYS) - set(config.keys())
-        extrakeys = set(config.keys()) - set(KEYS)
+    def validate(config, keystemp):
+        missingkeys = set(keystemp) - set(config.keys())
+        extrakeys = set(config.keys()) - set(keystemp)
```
standage, line 44: It makes sense to reuse this code for both the top-level configurations as well as the lower-level assembler configurations. However, it's poor design to outsource the coordination of the validation. Whatever code wants to validate the config has to validate the top-level config and then iterate over all the subconfigs and validate them separately. That code has to import the `KEYS1` and `KEYS2` constants as pass them as arguments. Why not hide all those details internally and give the user a simple interface?

yeat/cli/config.py
```diff
@@ -8,44 +8,40 @@
...
 
 
 class AssemblerConfig:
-    def __init__(self, algorithm, extra_args=""):
-        if algorithm not in ALGORITHMS:
-            raise AssemblyConfigurationError(f"Unsupported assembly algorithm '{algorithm}'")
-        self.algorithm = algorithm
-        self.extra_args = extra_args
-
-    @classmethod
```
standage, line 29: If you're no longer going to support creation of objects of type `AssemblerConfig`, then you don't need to define a class. These static methods can just be functions defined at the module level.

That said, I think there would be benefit to defining an `AssemblerConfig` class that stores a list of assemblers, a dictionary of samples, and provides functions for validating and accessing the config data.

It's no longer possible to create an object of type `AssemblerConfig`.

yeat/tests/test_config.py
```diff
@@ -7,44 +7,88 @@
...
+    assert assemblers == ["spades", "megahit", "unicycler"]
 
 
 def test_valid_config_entry():
-    data = {"algorithm": "spades", "extra_args": ""}
-    AssemblerConfig.validate(data)
+    data = json.load(open(data_file("configs/paired.cfg")))
+    AssemblerConfig.validate(data, KEYS1)
+    for assembler in data["assemblers"]:
+        AssemblerConfig.validate(assembler, KEYS2)
```
standage, line 56: The awkwardness of this code is what I was referring to earlier. Wouldn't something like this be clearer?

```python
config = AssemblerConfig.from_json(infile)
config.validate()
```

yeat/workflows/snakefiles/Pacbio
```diff
@@ -10,73 +10,82 @@
...
-        read=config["fastq"]
+        lambda wildcards: config["samples"][wildcards.sample]
     run:
-        if input.read.endswith(".gz"):
-            copyfile(input.read, output.read)
+        if input[0].endswith(".gz"):
+            copyfile(input[0], output.read)
         else:
-            shell(f"gzip -c {input.read} > {output.read}")
+            shell(f"gzip -c {input[0]} > {output.read}")
```
standage, line 37: This code silently ignores any filenames except for the first. Is the expectation of one file per sample for Pacbio data enforced anywhere else in the code?

yeat/workflows/workflows.py
```diff
@@ -7,39 +7,38 @@
...
-    r2 = Path(fastq2).resolve()
-    if not r2.is_file():
-        raise FileNotFoundError(f"No such file: '{r2}'")
-    assemblers = [config.algorithm for config in assembly_configs]
-    extra_args = {config.algorithm: config.extra_args for config in assembly_configs}
+    samples = {k: v.sample for k, v in assembly_samples.items()}
+    labels = [config.label for config in assembly_configs]
+    assemblers = {config.label: config.algorithm for config in assembly_configs}
+    extra_args = {config.label: config.extra_args for config in assembly_configs}
+    label_to_samples = {config.label: config.samples for config in assembly_configs}
```
standage, line 35: Seems like there's a lot of redundancy between this `run_paired` code and code in `run_pacbio`. That's a maintenance risk!

yeat/workflows/workflows.py
```diff
@@ -62,17 +61,22 @@ def check_canu_required_params(extra_args, cores):
...
-        raise FileNotFoundError(f"No such file: '{fastq}'")
-    assemblers = [config.algorithm for config in assembly_configs]
-    extra_args = {config.algorithm: config.extra_args for config in assembly_configs}
+    samples = {k: v.sample for k, v in assembly_samples.items()}
+    labels = [config.label for config in assembly_configs]
+    assemblers = {config.label: config.algorithm for config in assembly_configs}
+    extra_args = {config.label: config.extra_args for config in assembly_configs}
+    label_to_samples = {config.label: config.samples for config in assembly_configs}
     if "canu" in assemblers:
         check_canu_required_params(extra_args["canu"], cores)
```
standage, line 72: Seems like this is validation. Probably belongs with other config validation code?

yeat/workflows/workflows.py
```diff
@@ -7,39 +7,38 @@
 # Development Center.
 # -------------------------------------------------------------------------------------------------
```
standage, line 9: I'm not sure it makes sense to have an empty `yeat/workflows/__init__.py` file and then a separate `yeat/workflows/workflows.py` file. I'd suggest moving the contents of the latter into the former.

standage (changes requested):
I think the new config files are great. But I've pointed out a few places where I think the code could be improved a bit.

yeat/tests/test_cli.py
```diff
@@ -41,10 +38,15 @@ def test_invalid_custom_coverage_noninteger(coverage):
...
     with pytest.raises(ArgumentError, match=rf"{coverage} is not an integer"):
         args = cli.get_parser(exit_on_error=False).parse_args(arglist)
+
+
+@pytest.mark.parametrize(
+    "value",
+    [1, 10, 100],
+)
+def test_check_positive(value):
+    downsample.check_positive(value) == value
```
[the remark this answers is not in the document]
danejo3: Tests for bad input are tested in `test_invalid_custom_coverage_negative` and `test_invalid_custom_coverage_noninteger`.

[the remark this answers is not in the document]
standage: You're right: I missed that.

yeat/cli/config.py
```diff
@@ -8,44 +8,46 @@
...
-        if algorithm not in ALGORITHMS:
-            raise AssemblyConfigurationError(f"Unsupported assembly algorithm '{algorithm}'")
-        self.algorithm = algorithm
-        self.extra_args = extra_args
+    def __init__(self, data, threads):
+        self.data = data
+        self.validate()
+        self.threads = threads
+        self.create_sample_and_assembler_objects()
+        self.batch()
```
danejo3, line 32: The purpose of the `AssemblerConfig` class is to parse the "config" file and create `Sample` and `Assembler` objects based on the config's contents. When the config data is parsed, the data is first validated to ensure that the user entered in all the required entries. If correct, the class creates two list of classes of "samples" and "assemblers". Within each of those objects, further checking is done such as: Does the file exist? Did the user input enough "cores" for a given assembly algorithm? Is the assembly algorithm supported by YEAT?

Once all of the `Sample` and `Assembly` objects are created successfully, the `batch()` is called. The purpose of this function is to create a nested dictionary that are grouped by the algorithm in each `Assembly` object and has its corresponding `Sample` objects associated with them.

yeat/cli/config.py
```diff
@@ -54,3 +56,101 @@ def validate(config):
...
+        self.batch = {
+            "paired": {
+                "samples": self.get_batch_samples(paired_assemblers),
+                "assemblers": paired_assemblers,
+            },
+            "pacbio": {
+                "samples": self.get_batch_samples(pacbio_assemblers),
+                "assemblers": pacbio_assemblers,
+            },
+        }
```
danejo3, line 85: The result of the `batch()` is assigned in `self.batch`.

yeat/cli/config.py
```diff
@@ -54,3 +56,101 @@ def validate(config):
...
+            labels=[assembler.label for assembler in assemblers],
+            assemblers={assembler.label: assembler.algorithm for assembler in assemblers},
+            extra_args={assembler.label: assembler.extra_args for assembler in assemblers},
+            label_to_samples={assembler.label: assembler.samples for assembler in assemblers},
+            threads=args.threads,
+            downsample=args.downsample,
+            coverage=args.coverage,
+            genomesize=args.genome_size,
+            seed=args.seed,
+        )
```
danejo3, line 113: The purpose of this function is to convert the `Sample` and `Assembler` objects into non-class data nested dictionary. Snakemake can't take in classes in the config dictionary.

yeat/cli/config.py
danejo3: Updated the `validate()`. The function is called within the `AssemblerConfig` class `__init__()`.

yeat/cli/config.py
danejo3: Took your suggestions and definitely taking the benefits of the class in the new rewrite.

yeat/tests/test_config.py
danejo3: The `validate()` is now called in the `AssemblerConfig::__init__()`.

yeat/workflows/workflows.py
danejo3: Created the `to_dict()` in the `AssemblerConfig` object. Returns back a nested dictionary that each workflow uses: Paired, Pacbio, and Bandage.

yeat/workflows/workflows.py
danejo3: Moved this in the `Assembler` class object. This check happens as soon as the `Assembler` class is instantiated.

danejo3 (commented):
PR is ready for review! Refactored the `AssemblyConfig` object and integrated your suggestions! Tests were updated to reflect the new changes as well.

yeat/cli/config.py
```diff
@@ -54,3 +56,101 @@ def validate(config):
...
+                raise FileNotFoundError(f"No such file: '{read}'")
+            if str(read) in sample:
+                message = f"Found duplicate read sample: '{read}'"
+                raise AssemblyConfigurationError(message)
+            sample.append(str(read))
+        self.sample = sample
+
+    @classmethod
+    def from_data(cls, data):
+        return cls(data)
```
standage, line 131: This method isn't...doing anything. There's no difference between calling `Sample(data)` and calling `Sample.from_data(data)`.

yeat/cli/config.py
```diff
@@ -8,44 +8,46 @@
...
-        if len(algorithms) > len(set(algorithms)):
-            message = "Duplicate assembly configuration: please check config file"
+    def from_json(cls, infile, threads):
+        data = json.load(open(infile))
+        return cls(data, threads)
+
+    def validate(self):
+        self.check_keys(self.data, KEYS1)
+        for assembler in self.data["assemblers"]:
+            self.check_keys(assembler, KEYS2)
```
standage, line 42: The `KEYS1` and `KEYS2` variables are defined as constants at the top of the file but are used in only one place, for validation. I'd suggest it would be clearer to just declare them in-line in this code.

```python
        self.check_keys(self.data, ("samples", "assemblers"))
        for assembler in self.data["assemblers"]:
            self.check_keys(assembler, ("label", "algorithm", "extra_args", "samples"))
```

standage (commented):
Looks better! A couple more minor comments.

yeat/tests/test_paired.py
```diff
@@ -7,83 +7,87 @@
...
-    assert spades_result.exists()
-    megahit_result = Path(wd).resolve() / "analysis" / "megahit" / "final.contigs.fa"
-    assert megahit_result.exists()
+    analysis_dir = Path(wd).resolve() / "analysis"
+    expected = [
+        analysis_dir / "sample1" / "spades-default" / "spades" / "contigs.fasta",
+        analysis_dir / "sample1" / "spades-meta" / "spades" / "contigs.fasta",
+    ]
+    for observed in expected:
+        assert observed.exists()
```
standage, line 71: I use "observed" and "expected" as variable names frequently in test functions. I'm happy to see you picking that up, but I'm not sure the names are used appropriately in a couple cases here. Usually I use observed and expected to refer to the content or value of a variable.

```python
observed = some_function(test_data)
expected = 42
assert observed == expected
```

Here, you're referring to paths that you expect to exist. So the line `for observed in expected` is really misleading according to that pattern. More appropriate would be:

```python
expected_paths = [
    analysis_dir / "sample1" / "spades-default" / "spades" / "contigs.fasta",
    analysis_dir / "sample1" / "spades-meta" / "spades" / "contigs.fasta",
]
for expected in expected_paths:
    assert expected_paths.exists()
```

And honestly, this would be even more clear and concise.

```
assert (analysis_dir / "sample1" / "spades-default" / "spades" / "contigs.fasta").exists()
assert (analysis_dir / "sample1" / "spades-meta" / "spades" / "contigs.fasta").exists()
```

yeat/tests/test_paired.py
```diff
@@ -7,83 +7,87 @@
...
+    wd = str(tmp_path)
+    arglist = [
+        "--outdir",
+        wd,
+        data_file(f"configs/{algorithm}.cfg"),
+    ]
+    args = cli.get_parser().parse_args(arglist)
+    cli.main(args)
+    observed = Path(wd).resolve() / "analysis" / "sample1" / label / algorithm / expected
+    assert observed.exists()
```
standage, line 52: Same here: "expected" is more appropriate than "observed".

standage (approved):

danejo3:
Thanks for the code review!

PR #65, opened 2023-12-18
Adding grid support

danejo3:
The purpose of this PR is to integrate DRMAA grid support into YEAT.

Previously, YEAT would execute the snakemake workflow sequentially. To speed things up, parallel processing has been enabled. Users can adjust the grid execution parameters by passing them in YEAT's CLI. 

NOTE: Grid support has only been tested on SGE.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ testlinux:
 
 ## testgrid:    run only grid required automated tests
 testgrid:
-	pytest --cov=yeat -m grid --basetemp=yeat/tests/data/tmp
+	pytest --cov=yeat -m grid --basetemp=tmp
 
 ## testall:     run all tests but Bandage and linux required tests
 testall:
--- a/Makefile
+++ b/Makefile
@@ -35,9 +35,13 @@ testbandage:
 testlinux:
 	pytest --cov=yeat -m linux
 
+## testgrid:    run only grid required automated tests
+testgrid:
+	pytest --cov=yeat -m grid
+
 ## testall:     run all tests but Bandage and linux required tests
 testall:
-	pytest --cov=yeat -m 'not bandage and not linux'
+	pytest --cov=yeat -m 'not bandage and not linux and not grid'
 
 ## hifidata:    download PacBio HiFi-read test data for test suite
 hifidata:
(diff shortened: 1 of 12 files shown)


yeat/workflow/__init__.py
```diff
@@ -17,14 +17,28 @@ def run_workflow(args):
...
+            snakefile,
+            config=config,
+            cores=args.threads,
+            dryrun=args.dry_run,
+            printshellcmds=True,
+            workdir=args.outdir,
+            local_cores=args.threads,
+            nodes=args.grid_limit,
+            drmaa=setup_grid_args(args),
+            drmaa_log_dir=str((Path(args.outdir) / "gridlogs").resolve()),
```
danejo3, line 31: In order to save the DRMAA snakemake log files, this must be a full path and not relative.

Makefile
```diff
@@ -35,9 +35,13 @@ testbandage:
 testlinux:
 	pytest --cov=yeat -m linux
 
+## testgrid:    run only grid required automated tests
+testgrid:
+	pytest --cov=yeat -m grid --basetemp=tmp
+
```
danejo3, line 41: To run grid only tests,

> make testgrid

Grid tests are not included in `make testall`.

danejo3 (commented):
@standage 

This PR is ready for review! Thanks!

yeat/workflow/__init__.py
standage: Good to know!

standage (changes requested):
From a code review standpoint, this looks good. No concerns. I have not had a chance to test on the grid, but given the similarity to other workflows where we've implemented this approach, if you've had success with grid tests, I'm happy to accept that.

My only concern is your repeated claim that this is strictly SGE support. The whole point of DRMAA is that it's agnostic to the specific job scheduler used: SGE, PBS, Slurm, etc. It may be appropriate to say it's only been *tested* with SGE, but with DRMAA it should be straightforward to run on grids using a different scheduler. Please update the thread description, README, argument parser, etc., accordingly.

danejo3:
@standage Updated the wording in various requested documents. Let me know if I missed anything. Thanks!

standage (approved):

PR #68, opened 2024-01-10
Hybrid

danejo3:
The purpose of this PR is to introduce "hybrid" assembly to YEAT's workflow. To run hybrid mode, first, give a sample two types of reads: 1) pair-end, and 2) any kind of long read. Second, set an assembly's mode to "hybrid".

For example,

```
{
    "samples": {
        "Shigella_sonnei_53G": {
            "paired": [
                ["yeat/tests/data/short_reads_1.fastq.gz", "yeat/tests/data/short_reads_2.fastq.gz"]
            ],
            "pacbio-corr": [
                "yeat/tests/data/long_reads_high_depth.fastq.gz"
            ]
        }
    },
    "assemblies": {
        "unicycler-hybrid": {
            "algorithm": "unicycler",
            "extra_args": "",
            "samples": [
                "Shigella_sonnei_53G"
            ],
            "mode": "hybrid"
        }
    }
}
``` 

Below is a diagram of YEAT's output dir. In green are the final files created by YEAT. (Note: the `.done` file in the Bandage directory will exist if Bandage is on the user's computer.)

image


yeat/config/assembly.py
```diff
@@ -55,3 +62,47 @@ def check_canu_required_params(self):
...
+        elif self.mode == "hybrid" and (
+            sample.short_readtype != "paired" or sample.long_readtype not in LONG_READS
+        ):  # pragma: no cover
+            return False
+        return True
+
+    def check_sample_readtypes_match(self, check, sample):
+        if check == False:
+            message = f"No readtypes in '{sample.label}' match '{self.label}' assembly mode '{self.mode}'"
+            raise AssemblyConfigError(message)
```
danejo3, line 89: These three functions ensures that the samples associated with the assembly object has read types matching the assembly's mode.

For example,
Good - assembly mode: "paired" and sample's read type: "paired"
Bad - assembly mode: "paired" and sample's read type: "pacbio-hifi"

yeat/config/assembly.py
```diff
@@ -55,3 +62,47 @@ def check_canu_required_params(self):
...
+        algorithm_dir = f"analysis/{sample.label}/{readtype}/{self.label}/{self.algorithm}"
+        if self.bandage:
+            return f"{algorithm_dir}/bandage/.done"
+        return f"{algorithm_dir}/quast/report.html"
+
+    def get_readtype(self, sample):
+        if self.mode in ["paired", "single", "hybrid"]:
+            return self.mode
+        elif self.mode in ["pacbio", "oxford"]:  # pragma: no cover
+            return sample.long_readtype
```
danejo3, line 108: These three functions calculates the final expected file for each sample in the assembly's directory.

For example,
Assembly object has 3 samples: ["sample1", "sample2", and "sample3"].
User's computer does not have bandage.
Expected final files are:
- analysis/**sample1**/paired/default-spades/spades/quast/report.html
- analysis/**sample2**/paired/default-spades/spades/quast/report.html
- analysis/**sample3**/paired/default-spades/spades/quast/report.html

yeat/config/config.py
```diff
@@ -17,20 +17,20 @@
...
+        self.validate_config_keys()
         self.threads = threads
-        self.samples = {}
-        self.assemblies = {}
-        self.create_sample_and_assembly_objects()
-        self.validate_samples_to_assembly_modes()
+        self.bandage = bandage
+        self.samples = self.create_sample_objects()
+        self.assemblies = self.create_assembly_objects()
+        self.expected_files = self.get_expected_files()
```
danejo3, line 27: `self.expected_files` will have all the collated expected files from both sample and assembly objects.

yeat/config/config.py
```diff
@@ -48,23 +48,26 @@ def check_valid_keys(self, observed_keys, expected_keys):
...
-            if check == False:
-                message = f"No samples can interact with assembly mode '{assembly.mode}' for '{assembly_label}'"
-                raise AssemblyConfigError(message)
+    def get_expected_files(self):
+        expected_files = []
+        for sample in self.samples.values():
+            expected_files += sample.expected_files
+        for assembly in self.assemblies.values():
+            expected_files += assembly.expected_files
+        return expected_files
```
danejo3, line 73: This function loops through sample and assembly objects and gets all the calculated final files for snakemake input all.

yeat/config/config.py
```diff
@@ -48,23 +48,26 @@ def check_valid_keys(self, observed_keys, expected_keys):
...
-            self.samples[label] = Sample(label, sample)
+            samples[label] = Sample(label, sample)
+        return samples
+
+    def create_assembly_objects(self):
+        assemblies = {}
         for label, assembly in self.config["assemblies"].items():
-            self.assemblies[label] = Assembly(label, assembly, self.threads)
+            assembly["samples"] = self.get_sample_objects(assembly["samples"])
+            assemblies[label] = Assembly(label, assembly, self.threads, self.bandage)
```
danejo3, line 61: This here is new. To make the final file calculation possible in the Assembly class, we need to pass in their sample objects containing specific metadata such as read type and its quality.

Note: readtype is needed for the snakemake output directory structure. (see diagram in the PR's description.)

yeat/tests/__init__.py
```diff
@@ -54,35 +54,6 @@ def run_yeat(arglist):
...
-            elif assembly_obj["mode"] in ["pacbio", "oxford"]:
-                expected.append(
-                    analysis_dir
-                    / sample_label
-                    / next(iter(long_readtype))
-                    / assembly_label
-                    / assembly_obj["algorithm"]
-                    / FINAL_FILES[algorithm].replace("*", sample_label)
-                )
-    return expected
```
danejo3, line 84: This old implementation of calculating the expected final files for each combinatorial combination for samples and assemblies is not needed due to how we calculate them in the objects themselves now.

yeat/workflow/Hybrid
```diff
@@ -0,0 +1,28 @@
...
+# This software was prepared for the Department of Homeland Security (DHS) by the Battelle National
+# Biodefense Institute, LLC (BNBI) as part of contract HSHQDC-15-C-00064 to manage and operate the
+# National Biodefense Analysis and Countermeasures Center (NBACC), a Federally Funded Research and
+# Development Center.
+# -------------------------------------------------------------------------------------------------
+
+from yeat.workflow.aux import get_longread_file
+
+
+rule unicycler:
```
danejo3, line 13: Unicycler is the only algorithm to run hybrid assembly.

yeat/workflow/aux.py
```diff
@@ -81,66 +81,8 @@ def print_downsample_values(genome_size, avg_read_length, coverage, down, seed):
     print(f"[yeat] random seed for sampling: {seed}")
 
 
-def get_expected_files(config):
```
danejo3, line 84: From our paired programming, moved this calculation to the individual classes themself. Much cleaner now!

danejo3 (commented):
@standage Okay! This is ready for review! Thanks for the suggestion during our paired programming to break down an outside calculation to the individual classes themselves. Running the test suite all the way takes at least 4 hours to complete. There is a `make testhybrid` to run only the hybrid test.

yeat/tests/test_oxford.py
```diff
@@ -23,8 +23,8 @@ def test_oxford_nanopore_assemblers_dry_run(tmp_path):
 @pytest.mark.parametrize("algorithm", ["flye", "canu", "unicycler"])
 def test_oxford_nanopore_read_assemblers(algorithm, capsys, tmp_path):
     wd = str(tmp_path)
-    cores = str(get_core_count())
+    cores = get_core_count()
     config = write_config(algorithm, wd, "ont.cfg")
-    arglist = ["-o", wd, "-t", cores, config]
+    arglist = ["-o", wd, "-t", str(cores), config]
     run_yeat(arglist)
     expected_files_exist(wd, config, cores)
```
danejo3, line 30: `get_parser().parse_args()` can only accept a list of strings.  `cores` are casted into a `str` for this purpose.

`expected_files_exists` require `cores` to be an int. This variable type is important because if the algorithm is `canu`, the number of cores must be 4 or greater.

danejo3 (commented):
Found and fixed the bug! This is good to go!

yeat/config/config.py
standage: Looks good. But maybe could be even more concise?

```python
from itertools import chain

# ...

    def get_expected_files(self):
        expected_files = []
        for element in chain(self.samples.values(), self.assemblies.values()):
            expected_files += element.expected_files
        return expected_files
```

yeat/config/sample.py
```diff
@@ -7,7 +7,7 @@
 # Development Center.
 # -------------------------------------------------------------------------------------------------
 
-from . import ILLUMINA_READS, LONG_READS, AssemblyConfigError
+from . import *
```
standage, line 10: Please be explicit.

yeat/config/sample.py
```diff
@@ -85,3 +73,37 @@ def check_reads(self, reads):
                 message = f"Found duplicate read sample '{read}' for '{self.label}'"
                 raise AssemblyConfigError(message)
             self.all_reads.append(read)
+
+    def get_short_readtype(self):
+        illumina = set.intersection(set(self.sample.keys()), set(ILLUMINA_READS))
+        if len(illumina) > 1:
+            message = f"Max of 1 Illumina readtype per sample for '{self.label}'"
+            raise AssemblyConfigError(message)
+        return next(iter(illumina), None)
```
standage, line 82: Clever use of the default return value for the `next` function here. Would it make the intent clearer to add another case to the conditional?

```python
    def get_short_readtype(self):
        illumina = set.intersection(set(self.sample.keys()), set(ILLUMINA_READS))
        if len(illumina) > 1:
            message = f"Max of 1 Illumina readtype per sample for '{self.label}'"
            raise AssemblyConfigError(message)
        elif len(illumina) == 0:
            return None
        else:
            return next(iter(illumina))
```

Another thing to consider: a better name for the `illumina` variable?

yeat/config/sample.py
```diff
@@ -85,3 +73,37 @@ def check_reads(self, reads):
...
+
+    def get_long_readtype(self):
+        long = set.intersection(set(self.sample.keys()), set(LONG_READS))
+        if len(long) > 1:
+            message = f"Max of 1 long readtype per sample for '{self.label}'"
+            raise AssemblyConfigError(message)
+        return next(iter(long), None)
```
standage, line 89: Same suggestion with the `next` function here.

yeat/tests/__init__.py
standage: Awesome. I *love* deleting large blocks of code!

yeat/tests/test_hybrid.py
```diff
@@ -0,0 +1,29 @@
...
+# Copyright (c) 2024, DHS. This file is part of YEAT: http://github.com/bioforensics/yeat
+#
+# This software was prepared for the Department of Homeland Security (DHS) by the Battelle National
+# Biodefense Institute, LLC (BNBI) as part of contract HSHQDC-15-C-00064 to manage and operate the
+# National Biodefense Analysis and Countermeasures Center (NBACC), a Federally Funded Research and
+# Development Center.
+# -------------------------------------------------------------------------------------------------
+
+import pytest
+from yeat.tests import *
```
standage, line 11: xplct pls

yeat/tests/test_oxford.py
```diff
@@ -23,9 +23,8 @@ def test_oxford_nanopore_assemblers_dry_run(tmp_path):
 @pytest.mark.parametrize("algorithm", ["flye", "canu", "unicycler"])
 def test_oxford_nanopore_read_assemblers(algorithm, capsys, tmp_path):
     wd = str(tmp_path)
-    cores = str(get_core_count())
+    cores = get_core_count()
     config = write_config(algorithm, wd, "ont.cfg")
-    arglist = ["-o", wd, "-t", cores, config]
+    arglist = ["-o", wd, "-t", str(cores), config]
```
standage, line 28: Where is this `get_core_count()` function defined? Is there a way to limit or control this? If not, it could be problematic when running tests in CI or HPC environments.

yeat/workflow/Snakefile
```diff
@@ -61,3 +60,10 @@ module bandage_workflow:
     config: config
 
 use rule * from bandage_workflow as bandage_*
+
+
+module hybrid_workflow:
+    snakefile: "Hybrid"
+    config: config
+
+use rule * from hybrid_workflow as hybrid_*
```
standage, line 69: I think *this* use of star imports is fine for Snakefiles.

standage (changes requested):
This is looking great, @danejo3. Test passed on my machine. I made several comments that I hope will help make the code even more clean and clear.

yeat/tests/test_oxford.py
danejo3: The `get_core_count()` is located in `yeat/tests/__init__.py`. I created this function to get the core count of a system because these tests are incredibly long. I thought the more cores we can throw at it, the faster we can run these assemblies. We can use the default `core=1` parameter (make our assembly tests very slow) or we can hard cap it to 4. Up for any suggestions!

> Is there a way to limit or control this?

As of right now, we are not limiting or controlling this. We can control it if we want by setting a cap or limit.

```
import multiprocessing

def get_core_count():
    return multiprocessing.cpu_count()
```

> If not, it could be problematic when running tests in CI or HPC environments.

Yes, I do agree with you here.

yeat/config/assembly.py
danejo3: Really liked this suggestion. Especially for the final "else" statement. I was able to remove `# pragma: no cover` and improve test coverage.

yeat/config/config.py
danejo3: This is awesome! I totally forgot about `chain()`!

yeat/config/assembly.py
```diff
@@ -55,3 +62,47 @@ def check_canu_required_params(self):
...
+    def check_sample_readtypes_match(self, check, sample):
+        if check == False:
+            message = f"No readtypes in '{sample.label}' match '{self.label}' assembly mode '{self.mode}'"
+            raise AssemblyConfigError(message)
+
+    def get_expected_files(self):
+        expected_files = []
+        for sample in self.samples.values():
+            expected_files.append(self.get_qa_file(sample))
+        return expected_files
```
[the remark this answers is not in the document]
danejo3: Changed the term "expected_files" to "target_files". Was easy to update.

yeat/tests/test_oxford.py
danejo3: Will create an issue thread on this!

PR #73, opened 2024-03-15
Updating the test suite to reflect how Assembly objects are created

danejo3:
In this PR, #68, Assembly object requires a dictionary of samples objects instead of lists. This PR updates the test suite to reflect that change.

### DIFF
--- a/yeat/config/assembly.py
+++ b/yeat/config/assembly.py
@@ -80,7 +80,7 @@ class Assembly:
             return sample.long_readtype in OXFORD_READS
         elif self.mode == "hybrid":
             return sample.short_readtype == "paired" and sample.long_readtype in LONG_READS
-        else:
+        else:  # pragma: no cover
             message = f"Invalid assembly mode '{self.mode}'"
             raise AssemblyConfigError(message)
 
@@ -102,6 +102,6 @@ class Assembly:
             return self.mode
         elif self.mode in ["pacbio", "oxford"]:
             return sample.long_readtype
-        else:
+        else:  # pragma: no cover
             message = f"Invalid assembly mode '{self.mode}'"
             raise AssemblyConfigError(message)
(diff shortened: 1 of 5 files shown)


yeat/config/assembly.py
```diff
@@ -80,7 +80,7 @@ def mode_and_readtypes_are_compatible(self, sample):
             return sample.long_readtype in OXFORD_READS
         elif self.mode == "hybrid":
             return sample.short_readtype == "paired" and sample.long_readtype in LONG_READS
-        else:
+        else:  # pragma: no cover
             message = f"Invalid assembly mode '{self.mode}'"
             raise AssemblyConfigError(message)
```
danejo3, line 85: These blocks of "else" statements will never be executed because of all the check statements before hand. For example, in this function, we are checking if the mode and read types are matching. For readability, I avoided ending the if-blocks with an else statement. In PR #68, a suggestion was made using the else block to raise an error.  I liked this idea and I thought this would be helpful if for some reason or another, if we raise this error, this can be a sign to fix things upstream. Adding tests in the test suite for these else-blocks-that-will-never-be-executed is untestable. To increase pytest coverage, I added the `# pragma: no cover` for these short blocks of code.

danejo3 (commented):
@standage This PR is ready for review! I updated the test suite to reflect the new changes in the recently merged PR #68. I added a few comments. Let me know if you have any questions, concerns, or suggestions. Thanks!

yeat/config/assembly.py
standage: > To increase pytest coverage, I added the `# pragma: no cover` for these short blocks of code.

I think we should be clear that we add this pragma to block of code that, for whatever reason, we're deliberately *choosing* not to test. The increase in test coverage is a *result* of that choice, not the *motivation* for that choice. Your other reasoning is fine, I just want to make sure we're clear on that.

standage (approved):
LGTM, thanks!

PR #74, opened 2024-03-21
just-yeat-it: new YEAT CLI command

danejo3:
The purpose of this PR is to introduce a new CLI called `just-yeat-it` that executes a simple workflow of `yeat` with paired-end reads. This command does not require a config file and only one assembly algorithm can be run. By default, the selected algorithm is SPAdes but can be changed to MEGAHIT or Unicycler. Users can customize the assembly algorithm command line flags by using the `--extra-args` flag.

Below is the help output for `just-yeat-it`.

```
(yeat) FDT010LM-613660:yeat dane.jo$ just-yeat-it -h
usage: just-yeat-it [-h] [-n] [-o DIR] [-t T] [-l L] [-c C] [-d D] [-g G] [-s S] [--sample-label STR] [--assembly-label STR] [--megahit | --unicycler] [--extra-args STR] reads reads

required arguments:
  reads                 paired-end reads in FASTQ format

options:
  -h, --help            show this help message and exit

workflow configuration:
  -n, --dry-run         construct workflow DAG and print a summary but do not execute
  -o DIR, --outdir DIR  output directory; default is current working directory
  -t T, --threads T     number of available T threads for sequential and parallel processing jobs; by default, T=1

fastp configuration:
  -l L, --length-required L
                        discard reads shorter than the required L length after pre-preocessing; by default, L=50

downsampling configuration:
  -c C, --coverage C    target an average depth of coverage Cx when auto-downsampling; by default, C=150
  -d D, --downsample D  randomly sample D reads from the input rather than assembling the full set; set D=0 to perform auto-downsampling to a desired level of coverage (see --coverage); set D=-1 to disable downsampling; by default,
                        D=0
  -g G, --genome-size G
                        provide known genome size in base pairs (bp); by default, G=0
  -s S, --seed S        seed for the random number generator used for downsampling; by default, the seed is chosen randomly

sample configuration:
  --sample-label STR    set the sample label; by default, "sample1"

algorithm configuration:
  --assembly-label STR  set the assembly label; by default, "assembly1"
  --megahit             use MEGAHIT assembly algorithm; by default, SPAdes
  --unicycler           use Unicycler assembly algorithm; by default, SPAdes
  --extra-args STR      add assembly algorithm flags; for example, "--meta" or "--isolate --careful" for SPAdes; by default, empty string
```


danejo3 (commented):
@standage This PR is ready for review! Great idea with `just-yeat-it`! I'm open to any suggestions or thoughts on the changes made in this PR such as default names, CLI flags, etc. Thanks!

yeat/cli/just_yeat_it.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def algorithm_configuration(parser):
+    algorithm = parser.add_argument_group("algorithm configuration")
+    algorithm.add_argument(
+        "--assembly-label",
+        default="assembly1",
+        help='set the assembly label; by default, "assembly1"',
+        metavar="STR",
+    )
```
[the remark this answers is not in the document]
standage: I was going to suggest that we could eliminate the `--assembly-label` and `--sample-label` arguments altogether and just choose a generic label. But that would mean each `just-yeat-it` run would need its own working directory. If we allow the user to set the assembly and sample labels, we could have numerous runs writing to the same output directory. So yeah, let's leave this as is.

yeat/cli/just_yeat_it.py
```diff
@@ -0,0 +1,115 @@
...
+        action="store_const",
+        const=True,
+        help="use MEGAHIT assembly algorithm; by default, SPAdes",
+    )
+    group.add_argument(
+        "--unicycler",
+        action="store_const",
+        const=True,
+        help="use Unicycler assembly algorithm; by default, SPAdes",
+    )
```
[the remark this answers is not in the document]
standage: Instead of `--megahit` and `--unicycler` flags, I'd suggest an `--algorithm` flag that defaults to `spades` but can take `megahit` and `unicycler` as values.

standage (commented):
Code looks good, and the new interface works exactly as expected! One comment and one suggestion.

yeat/cli/just_yeat_it.py
```diff
@@ -0,0 +1,99 @@
...
+        default="assembly1",
+        help='set the assembly label; by default, "assembly1"',
+        metavar="STR",
+    )
+    algorithm.add_argument(
+        "--algorithm",
+        default="spades",
+        help='substitute the default assembly algorithm with another algorithm; for example, "megahit" or "unicycler"; by default, "spades"',
+        metavar="STR",
+    )
```
danejo3, line 61: Completely down to update the "help" tip if it's not clear or can be simpler.

yeat/cli/just_yeat_it.py
```diff
@@ -0,0 +1,99 @@
...
+    outfile = open(outdir / "config.cfg", "w")
+    json.dump(data, outfile, indent=4)
+
+
+def get_config_data(args):
+    return {
+        "samples": {args.sample_label: {"paired": [args.reads]}},
+        "assemblies": {
+            args.assembly_label: {
+                "algorithm": args.algorithm,
```
danejo3, line 88: With the `--algorithm` flag, we can directly plug in the input data into the config file.

yeat/tests/test_just_yeat_it.py
```diff
@@ -0,0 +1,76 @@
...
+        wd,
+        "--algorithm",
+        "INVALID",
+        data_file("short_reads_1.fastq.gz"),
+        data_file("short_reads_2.fastq.gz"),
+    ]
+    with pytest.raises(RuntimeError, match="Snakemake Failed"):
+        run_yeat(arglist)
+    out, err = capsys.readouterr()
+    assert "Invalid assembly algorithm 'INVALID' for 'assembly1'" in err
```
danejo3, line 76: If a user puts in a invalid `--algorithm`, the `Assembly` class will raise an error stating that the algorithm is invalid.

yeat/cli/just_yeat_it.py
```diff
@@ -0,0 +1,99 @@
...
+                "extra_args": args.extra_args,
+                "samples": [args.sample_label],
+                "mode": "paired",
+            }
+        },
+    }
```
danejo3, line 94: Bad inputs such as `args.algorithm` and `args.extra_args` will be caught in the `Assembly` class. For example, algorithms="BAD" will raise an error noting that the algorithm cannot be used with "paired-end" reads. Incomplete or missing "args.extra_args" will be caught as well. For example, `Canu` requires `genomesize={size}` and if not defined, `Canu` will error out.

danejo3 (commented):
Added suggestions! Ready for another PR review!

standage:
LGTM, thanks!

standage (approved):

PR #76, opened 2024-04-03
`yeat-auto` - automatically fill in samples for the config file

danejo3:
The purpose of this PR is to automatically fill in the configuration file. When users have multiple samples, for example, 10+, it can be tedious to fill out all the paths in the sample section of the config. To ease the manual labor and to prevent potential type-os, the `yeat-auto` command was created. When using this command, users will need to supply either a 1) sample name, 2) list of sample names, or 3) a text document containing all the sample names separated by newlines. The second required input is either a list of fastq files or a directory containing the sample's fastq files (see the `--seq-path` and `--files` flags).

Auto-populating is done by taking a sample's name and looking for matching fastq files based on substrings. 

For example, the sample name is `short_reads` and the matching files are `short_read_1.fastq.gz` and `short_read_2.fastq.gz`.

Auto-populating is only available for paired-end reads. As a result, all samples are dumped into the default spades algorithm in the "assemblies" section of the config.

```
"assemblies": {
    "spades-default": {
        "algorithm": "spades",
        "extra_args": "",
        "samples": [
            "short_reads"
        ],
        "mode": "paired"
    }
}
```

Users can then run the normal YEAT command with the newly created configuration, or if there are any changes that need to be made, users can adjust the config file as needed before passing it into `yeat`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,8 @@ setup(
     entry_points={
         'console_scripts': [
             'yeat=yeat.cli:main',
-            'just-yeat-it=yeat.cli.just_yeat_it:main'
+            'just-yeat-it=yeat.cli.just_yeat_it:main',
+            'yeat-auto=yeat.cli.yeat_auto:main'
         ],
     },
     classifiers=[
--- a/yeat/cli/auto_pop.py
+++ b/yeat/cli/auto_pop.py
@@ -23,6 +23,7 @@ class AutoPop:
     def __init__(self, args):
         self.samples = self.get_samples(args.samples)
         self.check_samples()
+        self.check_seq_path(args.seq_path)
         self.files = self.get_files(args.files, args.seq_path)
         self.check_files()
         self.files_to_samples = self.organize_files_to_samples()
@@ -44,6 +45,13 @@ class AutoPop:
                     message = f"cannot correctly process a sample name that is a substring of another sample name: {s1} vs. {s2}"
                     raise AutoPopError(message)
 
+    def check_seq_path(self, seq_path):
+        if not seq_path:
+            return
+        if not Path(seq_path).exists():
+            message = f"path does not exist: '{seq_path}'"
+            raise AutoPopError(message)
+
     def get_files(self, files, seq_path):
         if files:
             return [Path(f) for f in files]
@@ -55,7 +63,6 @@ class AutoPop:
         return files
 
     def check_files(self):
-        print(self.files)
         for file in self.files:
             if not file.exists():
                 message = f"file does not exist: {file}"
@@ -64,7 +71,7 @@ class AutoPop:
     def traverse(self, dirpath):
         dirpath = Path(dirpath)
         if not dirpath.is_dir():
-            return
+            return  # pragma: no cover
         for subpath in dirpath.iterdir():
             if subpath.is_dir():
                 yield from self.traverse(subpath)
--- a/yeat/cli/auto_pop.py
+++ b/yeat/cli/auto_pop.py
@@ -78,11 +78,22 @@ class AutoPop:
                 raise AutoPopError(message)
 
     def write_config_file(self, outfile):
-        samples = {}
-        for label, reads in self.files_to_samples.items():
-            samples[label] = {"paired": [reads]}
-        data = {"samples": samples, "assemblies": {}}
+        data = self.get_config_data()
         outdir = Path(outfile).parent
         outdir.mkdir(parents=True, exist_ok=True)
         outfile = open(outfile, "w")
         json.dump(data, outfile, indent=4)
+
+    def get_config_data(self):
+        samples = {}
+        for label, reads in self.files_to_samples.items():
+            samples[label] = {"paired": [reads]}
+        assemblies = {
+            "spades-default": {
+                "algorithm": "spades",
+                "extra_args": "",
+                "samples": self.samples,
+                "mode": "paired",
+            }
+        }
+        return {"samples": samples, "assemblies": assemblies}
--- a/yeat/cli/auto_pop.py
+++ b/yeat/cli/auto_pop.py
@@ -41,7 +41,7 @@ class AutoPop:
                     continue
                 if s1 in s2:
                     message = f"cannot correctly process a sample name that is a substring of another sample name: {s1} vs. {s2}"
-                    raise ValueError(message)
+                    raise AutoPopError(message)
 
     def get_files(self, files, seq_path):
         if files:
@@ -82,7 +82,7 @@ class AutoPop:
         for label, reads in self.files_to_samples.items():
             samples[label] = {"paired": [reads]}
         data = {"samples": samples, "assemblies": {}}
-        outdir = Path(outfile).parent.absolute()
+        outdir = Path(outfile).parent
         outdir.mkdir(parents=True, exist_ok=True)
         outfile = open(outfile, "w")
         json.dump(data, outfile, indent=4)
--- a/yeat/cli/auto_pop.py
+++ b/yeat/cli/auto_pop.py
@@ -1,0 +1,88 @@
+# -------------------------------------------------------------------------------------------------
+# Copyright (c) 2024, DHS. This file is part of YEAT: http://github.com/bioforensics/yeat
+#
+# This software was prepared for the Department of Homeland Security (DHS) by the Battelle National
+# Biodefense Institute, LLC (BNBI) as part of contract HSHQDC-15-C-00064 to manage and operate the
+# National Biodefense Analysis and Countermeasures Center (NBACC), a Federally Funded Research and
+# Development Center.
+# -------------------------------------------------------------------------------------------------
+
+from collections import defaultdict
+import json
+from pathlib import Path
+
+
+EXTENSIONS = (".fastq", ".fastq.gz", ".fq", ".fq.gz")
+
+
+class AutoPopError(ValueError):
+    pass
+
+
+class AutoPop:
+    def __init__(self, args):
+        self.samples = self.get_samples(args.samples)
+        self.check_samples()
+        self.files = self.get_files(args.files, args.seq_path)
+        self.files_to_samples = self.organize_files_to_samples()
+        self.check_files_to_samples()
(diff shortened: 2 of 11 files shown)


setup.py
```diff
@@ -34,7 +34,8 @@
     entry_points={
         'console_scripts': [
             'yeat=yeat.cli:main',
-            'just-yeat-it=yeat.cli.just_yeat_it:main'
+            'just-yeat-it=yeat.cli.just_yeat_it:main',
+            'yeat-auto=yeat.cli.yeat_auto:main'
```
danejo3, line 38: New command: `yeat-auto`

yeat/cli/auto_pop.py
```diff
@@ -0,0 +1,112 @@
...
+class AutoPop:
+    def __init__(self, args):
+        self.samples = self.get_samples(args.samples)
+        self.check_samples()
+        self.check_seq_path(args.seq_path)
+        self.files = self.get_files(args.files, args.seq_path)
+        self.check_files()
+        self.files_to_samples = self.organize_files_to_samples()
+        self.check_files_to_samples()
+        self.write_config_file(args.outfile)
```
danejo3, line 31: Created the `AutoPop` class to create the config file. Samples and files are parsed from the command-line arguments and checked for validity. If input data has no problems, we sort the fastq files to their samples and do another check to ensure that there are enough reads per sample. If sorted correctly, we then write the config file.

yeat/cli/auto_pop.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def traverse(self, dirpath):
+        dirpath = Path(dirpath)
+        if not dirpath.is_dir():
+            return  # pragma: no cover
+        for subpath in dirpath.iterdir():
+            if subpath.is_dir():
+                yield from self.traverse(subpath)
+            else:
+                yield subpath
```
danejo3, line 79: `--seq-path` can take in a nested directory to find fastq files.

yeat/cli/auto_pop.py
```diff
@@ -0,0 +1,112 @@
...
+        for label, reads in self.files_to_samples.items():
+            samples[label] = {"paired": [[str(read.absolute()) for read in reads]]}
+        assemblies = {
+            "spades-default": {
+                "algorithm": "spades",
+                "extra_args": "",
+                "samples": self.samples,
+                "mode": "paired",
+            }
+        }
```
danejo3, line 111: All samples will be dumped into the `spades-default` assembly since the auto-populate feature is only for paired-end reads.

yeat/cli/yeat_auto.py
```diff
@@ -0,0 +1,65 @@
...
+        help="path to a directory containing FASTQ files to use as input; incompatible with --files",
+        metavar="PATH",
+    )
+    meg.add_argument(
+        "--files",
+        default=None,
+        help="a list of FASTQ files to use as input; incompatible with --seq-path",
+        metavar="FQ",
+        nargs="+",
+    )
```
danejo3, line 56: Users will need to provide either one of these flags along with the sample or list of sample names.

danejo3 (commented):
This PR is ready for review! @standage

yeat/cli/auto_pop.py
```diff
@@ -0,0 +1,112 @@
...
+
+class AutoPopError(ValueError):
+    pass
+
+
+class AutoPop:
+    def __init__(self, args):
+        self.samples = self.get_samples(args.samples)
+        self.check_samples()
+        self.check_seq_path(args.seq_path)
```
standage, line 26: The `args` argument here hides all the data that the `AutoPop` object depends on to work properly. If we explicitly pass that data in discrete arguments, it will more clearly communicate what the object needs and what the code is doing.

It looks like there are four arguments: samples, files, seq_path, and outfile. "Clean Code" purists would argue that this is too many, and in some cases they may be right. But I'm not sure I see a way to implement this feature without passing at least the first three to the same object. You could definitely simplify by moving the `write_config_file` method outside of the constructor, requiring the user to call it explicitly, and passing `outfile` as an argument to that method.

yeat/cli/auto_pop.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def get_files(self, files, seq_path):
+        if files:
+            return [Path(f) for f in files]
+        files = []
+        for file in self.traverse(seq_path):
+            if not file.name.endswith(EXTENSIONS):
+                continue
+            files.append(file)
+        return files
```
standage, line 63: This method violates the "do one thing" rule. Perhaps it would be better to remove the first two lines, drop the `files` argument, and call the method like this.

```
self.files = files if files is not None else self.get_files(seq_path)
```

yeat/tests/test_yeat_auto.py
```diff
@@ -0,0 +1,124 @@
...
+        run_yeat_auto(arglist)
+
+
+def test_sequence_with_no_files(tmp_path):
+    wd = str(tmp_path)
+    config = f"{wd}/config.cfg"
+    arglist = ["SAMPLE", "-o", config, "--seq-path", data_file("")]
+    message = "sample SAMPLE: expected 2 FASTQ files for paired-end data, found 0"
+    with pytest.raises(AutoPopError, match=message):
+        run_yeat_auto(arglist)
```
standage, line 69: Maybe change `SAMPLE` here to something that's more obviously bogus.

yeat/cli/auto_pop.py
standage, line 1: I don't have *super* strong feelings about this, but I think one could argue that this class is out of scope for the CLI and could be moved up a level of organization to the main YEAT package or the `config` subpackage.

yeat/cli/yeat_auto.py
standage, line 1: Argument groups become very helpful when a program has a large number of arguments and options. For a program this size with essentially four arguments, breaking things down that granularly isn't as effective in my opinion.

```
usage: yeat-auto [-h] [-o FILE] (--seq-path PATH | --files FQ [FQ ...]) samples [samples ...]

required arguments:
  samples               list of sample names or path to .txt file containing sample names

options:
  -h, --help            show this help message and exit

config configuration:
  -o FILE, --outfile FILE
                        output config file; by default, "config.cfg"

input configuration:
  --seq-path PATH       path to a directory containing FASTQ files to use as input; incompatible with --files
  --files FQ [FQ ...]   a list of FASTQ files to use as input; incompatible with --seq-path
```

I think the example below is more effective and aesthetically pleasing than the example above.

```
usage: yeat-auto [-h] [-o FILE] (--seq-path PATH | --files FQ [FQ ...]) samples [samples ...]

required arguments:
  samples               list of sample names or path to .txt file containing sample names

options:
  -h, --help            show this help message and exit
  --seq-path PATH       path to a directory containing FASTQ files to use as input; incompatible with --files
  --files FQ [FQ ...]   a list of FASTQ files to use as input; incompatible with --seq-path
  -o FILE, --outfile FILE
                        output config file; by default, "config.cfg"
```

I also think printing the config to the standard output is a more conventional default than a specific filename.

standage (changes requested):
The new program works as expected in my hands, and the code looks great! I've made a few suggestions.

danejo3 (commented):
Updated code with suggestions! Ready for another review!

yeat/config/auto_pop.py
```diff
@@ -0,0 +1,109 @@
...
+                if s1 in s2:
+                    message = f"cannot correctly process a sample name that is a substring of another sample name: {s1} vs. {s2}"
+                    raise AutoPopError(message)
+
+    def check_seq_path(self, seq_path):
+        if not seq_path:
+            return
+        if not Path(seq_path).exists():
+            message = f"path does not exist: '{seq_path}'"
+            raise AutoPopError(message)
```
standage, line 52: There's a built-in exception type for this use case.

```python
    def check_seq_path(self, seq_path):
        if not seq_path:
            return
        if not Path(seq_path).exists():
            raise FileNotFoundError(seq_path)
```

yeat/config/auto_pop.py
```diff
@@ -0,0 +1,109 @@
...
+        for file in self.files:
+            if not file.exists():
+                message = f"file does not exist: {file}"
+                raise AutoPopError(message)
+
+    def organize_files_to_samples(self):
+        files_to_samples = defaultdict(list)
+        for sample in self.samples:
+            files_to_samples[sample] = [file for file in self.files if sample in str(file)]
+        return files_to_samples
```
standage, line 82: You could get away with just declaring a `dict` here. The `defaultdict` only provides an advantage if you want to call `dict_of_lists[key].append()` without first calling `dict_of_lists[key] = []`.

yeat/cli/yeat_auto.py
```diff
@@ -0,0 +1,60 @@
...
+    parser.add_argument(
+        "-o",
+        "--outfile",
+        default="config.cfg",
+        help='output config file; by default, "config.cfg"',
+        metavar="FILE",
+        type=str,
+    )
```
standage, line 51: Reiterating my suggestion that stdout is used as default.

```python
    parser.add_argument(
        "-o",
        "--outfile",
        default=sys.stdout,
        help='output config file; by default, output is written to terminal (standard output)',
        metavar="FILE",
        type=argparse.FileType("w"),
    )
```

The `.write_config_file()` method would need to be updated to take a file handle instead of a string, and the main method to do the directory handling, but that's probably more appropriate anyway given the scope of these units of code.

standage (changes requested):
A few more minor changes.

yeat/config/auto_pop.py
danejo3: Great catch! I updated another function that did the same thing as well.

yeat/config/auto_pop.py
danejo3: Good point 👍

yeat/cli/yeat_auto.py
danejo3: Completely missed the previous comment. I agree that the stdout is the way to go. Simplified the code significantly and users can pipe the output into a file, etc.

danejo3:
Comments addressed! Thanks!

PR #77, opened 2024-04-15
Customize downsample, depth of coverage, and genome size per sample

### LINKED ISSUE #75: Providing genome size for each sample and clarify downsample flag
YEAT has flags to customize the downsampling step. One of the flags that can be used is the `--genome-size` flag. When a config file has multiple samples, users cannot apply the known genome size of each sample. Instead, the provided genome size is applied to all samples. This can be problematic if the known sizes are different for each sample. One way to resolve this is by adding a `genome-size` key for each sample in the config.

Continuing the conversation with genome size, the help message in YEAT for `--genome-size` and `--downsample` needs further clarification. If a user provides a sample's genome size, downsampling will need to be calculated. However, if downsampling is provided with the `--downsample` flag, there is no need to use the calculated genome size from `mash` and recalculate the down. This logic is coded correctly in the `Shared` snakemake file under rule `downsample`. In YEAT's help message, this idea is not clear and probably should state somewhere that users shouldn't use both the `--downsample` and `--genome-size` flags because `--downsample` will disregard `--genome-size`.

danejo3:
The purpose of this MR is to resolve #75.

Previously, YEAT would adjust downsample configuration values based on the CLI inputs and applied to all samples noted in the config file. This can be problematic if two or more samples are not the same. In this MR, we added specific entries for each sample to take in `downsample`, `genome_size`, and `coverage_depth` for paired and single-end reads.  Just like the CLI flags, users will add the same values to the config file for each sample.

Because `just-yeat-it` does not use a config file, the original downsample configuration flags are still available.

The only flag that still remains for `yeat` is `--seed`.

Below is an example of a config file with the following updates:

```
{
    "samples": {
        "sample1": {
            "paired": [
                [
                    "yeat/tests/data/short_reads_1.fastq.gz",
                    "yeat/tests/data/short_reads_2.fastq.gz"
                ]
            ],
            "downsample": 0,
            "genome_size": 0,
            "coverage_depth": 150
        },
        "sample2": {
            "paired": [
                [
                    "yeat/tests/data/Animal_289_R1.fq.gz",
                    "yeat/tests/data/Animal_289_R2.fq.gz"
                ]
            ],
            "downsample": 0,
            "genome_size": 0,
            "coverage_depth": 150
        },
        "sample3": {
            "pacbio-hifi": [
                "yeat/tests/data/ecoli.fastq.gz"
            ],
            "downsample": 0,
            "genome_size": 0,
            "coverage_depth": 150
        },
        "sample4": {
            "nano-hq": [
                "yeat/tests/data/ecolk12mg1655_R10_3_guppy_345_HAC.fastq.gz"
            ],
            "downsample": 0,
            "genome_size": 0,
            "coverage_depth": 150
        }
    },
    "assemblies": {
        "spades-default": {
            "algorithm": "spades",
            "extra_args": "",
            "samples": [
                "sample1",
                "sample2"
            ],
            "mode": "paired"
        },
        "hicanu": {
            "algorithm": "canu",
            "extra_args": "genomeSize=4.8m",
            "samples": [
                "sample3"
            ],
            "mode": "pacbio"
        },
        "flye_ONT": {
            "algorithm": "flye",
            "extra_args": "",
            "samples": [
                "sample4"
            ],
            "mode": "oxford"
        }
    }
}
```


yeat/tests/test_config.py
```diff
@@ -9,46 +9,13 @@
...
 from yeat.config import AssemblyConfigError
-from yeat.config.assembly import Assembly
 from yeat.config.config import AssemblyConfig
-from yeat.config.sample import Sample
 from yeat.tests import data_file
 
 pytestmark = pytest.mark.short
 
 
-def test_invalid_assembly_algorithm():
```
danejo3, line 22: Moved tests to different modules: `test_sample.py` and `test_assembly.py` for organization. Keeping `test_config.py` to contain tests testing for the `AssemblyConfig` class object.

danejo3 (commented):
This MR is ready for review! @standage A lot of file changes; however, most changes are adding `downsample`, `genome_size`, and `coverage_depth` to the json data. After this merge, we'll cut a new release for YEAT!

yeat/workflow/__init__.py
```diff
@@ -51,8 +52,8 @@ def run_workflow(args):
 def resolve_paths(infile):
     data = json.load(open(infile))
     for sample in data["samples"].values():
-        for readtype, reads in sample.items():
-            sample[readtype] = get_resolved_paths(reads)
+        for key, value in sample.items():
+            sample[key] = value if key in DOWNSAMPLE_KEYS else get_resolved_paths(value)
```
standage, line 56: This code was updated, but it still needs some work. Changing `readtype, reads` to `key, value` reflects the wider scope of the function, but it also makes the code harder to understand. Also, the name of the function suggests that it's working only with paths, but that's no longer true.

standage (changes requested):
Looks good! As you said, there were lots of little updates to account for the new config structure, but the primary code changes were clear and straightforward to follow.

There's one function that I don't think fared very well in this update: see below.

yeat/workflow/__init__.py
```diff
@@ -49,14 +49,19 @@ def run_workflow(args):
...
+        resolve_sample_paths(sample)
     return data
 
 
+def resolve_sample_paths(sample):
+    readtypes = set(sample.keys()).intersection(set(READ_TYPES))
+    for readtype in readtypes:
+        sample[readtype] = get_resolved_paths(sample[readtype])
+
+
```
danejo3, line 64: Thoughts on the newly refactored function? @standage Other than that, this is PR is ready!

standage (approved):

PR #83, opened 2024-11-29
Consolidating the README

danejo3:
The purpose of this PR is to consolidate and simplify the README.

Previously, the README contained various notes that users needed to be aware of regarding installation, usage, and other aspects. To improve organization, a dedicated wiki page has been created to better structure these helpful notes.

https://github.com/bioforensics/yeat/wiki

While consolidating, I have made a few changes to make the installation process easier for Linux users. Packages that had to be manually installed by users such as `Bandage` and `metaMDBG` are now available on conda and can be installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,52 +1,9 @@
 # YEAT
 
-YEAT (**Y**our **E**verday **A**ssembly **T**ool) is an all-in-one platform designed to assemble multiple samples of varying read types (both short and long) using a combination of reliable, widely used, and cutting-edge algorithms such as SPAdes and Flye. It utilizes a Snakemake workflow to preprocess sample reads, perform assembly, and postprocess the resulting contigs in parallel.
+## Intro
 
-| Read Types  | Supported Algorithms |
-| ------------- | ------------- |
-| Paired-end  | SPAdes<sup>[1](#reference-1)</sup>, MEGAHIT<sup>[2](#reference-2)</sup>, Unicycler<sup>[3](#reference-3)</sup>, PenguiN<sup>[4](#reference-4)</sup>, VelvetOptimiser<sup>[5](#reference-5)</sup> |
-| Single-end | SPAdes<sup>[1](#reference-1)</sup>, MEGAHIT<sup>[2](#reference-2)</sup>, Unicycler<sup>[3](#reference-3)</sup>, PenguiN<sup>[4](#reference-4)</sup>, VelvetOptimiser<sup>[5](#reference-5)</sup> |
-| PacBio CLR (<20% error) | Flye<sup>[6](#reference-6)</sup>, Canu<sup>[7](#reference-7)</sup>, Unicycler<sup>[3](#reference-3)</sup> |
-| PacBio Corrected (<3% error) | Flye<sup>[6](#reference-6)</sup>, Canu<sup>[7](#reference-7)</sup>, Unicycler<sup>[3](#reference-3)</sup> |
-| PacBio HiFi (<1% error) | Flye<sup>[6](#reference-6)</sup>, Canu<sup>[7](#reference-7)</sup>, Hifiasm<sup>[8](#reference-8)</sup>, Hifiasm-meta<sup>[9](#reference-9)</sup>, Unicycler<sup>[3](#reference-3)</sup>, metaMDBG<sup>[10](#reference-10)</sup> |
-| ONT Regular (<20% error) | Flye<sup>[6](#reference-6)</sup>, Canu<sup>[7](#reference-7)</sup>, Unicycler<sup>[3](#reference-3)</sup> |
-| ONT Corrected (<3% error) | Flye<sup>[6](#reference-6)</sup>, Canu<sup>[7](#reference-7)</sup>, Unicycler<sup>[3](#reference-3)</sup> |
-| ONT High-quality (<5% error) | Flye<sup>[6](#reference-6)</sup>, Canu<sup>[7](#reference-7)</sup>, Unicycler<sup>[3](#reference-3)</sup> |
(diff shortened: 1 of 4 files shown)


danejo3 (commented):
This PR is ready for review, and I'd love to hear any comments or feedback! Moving many of the installation instructions and other notes to the wiki has really helped clean up the README. I'm still working on the wiki and figuring out the best way to organize it, so I'd appreciate any suggestions for improvement.

A key change in this PR is streamlining the installation process for Linux users. Originally, I had omitted certain packages in the `environment.yml` file that were unavailable on macOS. I’ve now updated the file to focus specifically on Linux users. If there are packages that macOS users need to install, they can find detailed information and instructions on the wiki.

README.md
```diff
@@ -1,257 +1,52 @@
...
+| PacBio CLR (<20% error) | Flye<sup>[6](#reference-6)</sup>, Canu<sup>[7](#reference-7)</sup>, Unicycler<sup>[3](#reference-3)</sup> |
+| PacBio Corrected (<3% error) | Flye<sup>[6](#reference-6)</sup>, Canu<sup>[7](#reference-7)</sup>, Unicycler<sup>[3](#reference-3)</sup> |
+| PacBio HiFi (<1% error) | Flye<sup>[6](#reference-6)</sup>, Canu<sup>[7](#reference-7)</sup>, Hifiasm<sup>[8](#reference-8)</sup>, Hifiasm-meta<sup>[9](#reference-9)</sup>, Unicycler<sup>[3](#reference-3)</sup>, metaMDBG<sup>[10](#reference-10)</sup> |
+| ONT Regular (<20% error) | Flye<sup>[6](#reference-6)</sup>, Canu<sup>[7](#reference-7)</sup>, Unicycler<sup>[3](#reference-3)</sup> |
+| ONT Corrected (<3% error) | Flye<sup>[6](#reference-6)</sup>, Canu<sup>[7](#reference-7)</sup>, Unicycler<sup>[3](#reference-3)</sup> |
+| ONT High-quality (<5% error) | Flye<sup>[6](#reference-6)</sup>, Canu<sup>[7](#reference-7)</sup>, Unicycler<sup>[3](#reference-3)</sup> |
+
+## Installation (for Linux only)
```
standage, line 16: Here, the entire installation instructions are described as for Linux only. In the Wiki, only the metaMDBG and velvet installs are listed as Linux-only. A few points to consider:

- I think the approach you've begun in the Wiki is probably better. Point out which installation instructions are platform-specific, and which will work across the board.
- Is velvet-optimiser really not compatible with macOS? Both the README and the Wiki imply this, but I don't think that's the case. If it's not, you need to be careful the documentation doesn't imply something that's not true.
- The second point here highlights a common concern: when code or documentation is duplicated, it can easily get out of sync. You have to be disciplined to remember (or use a checklist to be reminded) to do the same update in multiple places. You might consider keeping the README more minimal if you want to include its current contents on the wiki.

README.md
```diff
@@ -1,257 +1,52 @@
...
-In order to run [`Bandage`](https://github.com/rrwick/Bandage) in YEAT's workflow, users will need to manually download the pre-built binaries files [here](https://rrwick.github.io/Bandage/) or build it from the source code [here](https://github.com/rrwick/Bandage). Instructions for building Bandage from the source code can be found [here](https://github.com/rrwick/Bandage#building-from-source).
-
-Once the binary file has been obtained, users will need to update their environment's `$PATH` to include the path to the directory containing the binary file.
+## Usage
 
-Example for Ubuntu:
 ```
-export PATH=~/projects/Bandage:$PATH
+yeat-auto short_reads --files {short_read1} {short_read2} > config.cfg
+yeat --outdir {path} config.cfg
```
standage, line 32: You might want to make it clear that the user should manually check edit the config file between running `yeat-auto` and `yeat`. And I'd guess that the `--seq-path` option would be the more common configuration route.

```
yeat-auto {sample names} --seq-path {path to dir with Fastq files} > config.cfg
# Check and edit config.cfg before proceeding
yeat --outdir {path} config.cfg
```

environment.yml
```diff
@@ -31,3 +31,4 @@ dependencies:
     - samtools>=1.21
     - bowtie2>=2.5
     - pilon>=1.24
+    - bandage>=0.8
```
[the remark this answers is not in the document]
standage: If this is the reason you've provided Linux-only instructions, maybe you can take this out and add another couple line to the instructions to make them cross-platform again.

```
# Linux only; Bandage must be installed manually on macOS
conda install -c bioconda 'bandage>=0.8'
```

standage (changes requested):
Looking good! A few comments and suggestions.

yeat/workflow/Pacbio
```diff
@@ -98,8 +98,6 @@ rule metamdbg:
         contigs="analysis/{sample}/{readtype,pacbio-raw|pacbio-corr|pacbio-hifi}/{label}/metamdbg/contigs.fasta"
     input:
         reads="seq/input/{sample}/{readtype}/combined-reads.fq.gz"
-    conda:
-        "yeat-metaMDBG"
```
danejo3, line 102: `metaMDBG` can now be installed with conda. No need to create a new dedicated conda environment.

README.md
```diff
@@ -1,257 +1,9 @@
...
-            "extra_args": "",
-            "samples": [
-                "sample4"
-            ],
-            "mode": "oxford"
-        }
-    }
-}
-```
+YEAT documentation is available on the [YEAT GitHub wiki](https://github.com/rrwick/Bandage/wiki).
```
danejo3, line 9: Made the README significantly shorter and simpler. Added a link to the GitHub wiki page for installation instructions, settings, etc.

danejo3 (commented):
Thanks for the suggestions! @standage Simplified the README even further. Previous suggestions that were made were added to the wiki. After this PR, I have one more PR that I would like to add before cutting a release to accommodate the new updates for `metaMDBG`.

README.md
danejo3: Moved the table and installation instructions to wiki to reduce the challenges of updating multiple places. I made the decision to move the table to the wiki because new algorithms will be added to the table in the future.

README.md
standage: Looks good. I'd just recommend pointing to YEAT's wiki instead of Bandage's wiki. 🙃

standage (changes requested):
Wiki and README look good! Just one link to be fixed.

standage (approved):

PR #95, opened 2025-08-27
Adding large test files with GitLFS

danejo3:
Originally, we stored the large test data in OSF and used a make command to download them into YEAT's test directory. We are moving away from that approach and instead placing the test files directly in the repository. These files are relatively small and can utilize GitHub's LFS data management system.


danejo3:
@standage This MR is ready for review!

PR #102, opened 2025-10-02
Myloasm

danejo3:
The purpose of this PR is to introduce and bring in a new assembler: myloasm. Myloasm is a new high-resolution long-read metagenome assembler. Internal testing has shown that this assembler does pretty well with metagenomic samples.


danejo3:
New instructions for myloasm for MacOS 

```
# Install myloasm from the source into the yeat environment
git clone https://github.com/bluenote-1577/myloasm --recurse-submodules  
cd myloasm
cargo build --release
cp target/release/myloasm $CONDA_PREFIX/bin/
```

Issue #108, opened 2026-02-27
Trimming Pacbio HIFI long reads


danejo3:
During internal testing, we identified cases in which PacBio HIFI reads required trimming prior to downstream processing. Once our trimming tool is released on GitHub, we should integrate it into YEAT. Currently, HIFI reads does not undergo filtering or trimming.

PR #109, opened 2026-03-20
Improving downsampling strategy

danejo3:
The purpose of this PR is to add the ability to downsample with BBNorm for Illumina reads. To do this, we added the "method" key in the TOML file to allow users to switch from "seqtk", "bbnorm", or skip downsampling all together. However, in order to do this, various other changes were needed to make this change.

Here are the major changes related to this structural change:
1. Reorganized and expanded the TOML structure to make it more intuitive and user-friendly.
2. Refactored the configuration classes, introducing a Config class with supporting subclasses like FilterSettings and DownsampleSettings.
3. Added and revised several TOML keys, along with updated default values.

Here's the new TOML file format example.

```
[global_settings.filter.short]
enabled = false
fastp_args = ""

[global_settings.filter.long]
enabled = false
chopper_args = ""

[global_settings.downsample.short]
enabled = false
method = "random"                           # random|bbnorm
target_depth = 150                          # desired depth of coverage
target_num_reads = "auto"                   # number or "auto" to calculate from target_depth and genome_size; ignored if mode = "bbnorm"
genome_size = "auto"                        # number (if genome size known) or "auto" to estimate using Mash; ignored if mode = "bbnorm"

[global_settings.downsample.long]
enabled = false
target_depth = 150                          # desired depth of coverage
target_num_reads = "auto"                   # number or "auto" to calculate from target_depth and genome_size
genome_size = "auto"                        # number (if genome size known) or "auto" to estimate using Mash

[samples.sample1]
illumina = "data/short_reads_R?.fastq.gz"   # illumina|pacbio_hifi|ont_simplex|ont_duplex|ont_ultralong

[samples.sample2]
illumina = "data/short_reads_R?.fastq.gz"
pacbio_hifi = "data/long_reads.fastq.gz"

[assemblers.assembly1]
algorithm = "spades"
arguments = ""
```


danejo3:
2 config examples that will be shown when executing `yeat --init`. The first one is the current one with slight changes to required key names.

```
[global_settings]
# read filtering parameters
filter = true                           # if false, proceed to downsample
min_length = 100                        # adjust for long reads   
quality = 15                            # adjust for long reads  

# downsampling parameters
downsampling = "none"                   # if downsampling="none", proceed to assembly
target_num_reads = 0                    # applies when downsampling="random"
genome_size = 0                         # applies when downsampling="random" and target_num_reads=0
target_depth = 150                      # applies when downsampling="random" and target_num_reads=0


[samples.sample1]
illumina = "short_reads_?.fastq.gz"     # glob path

[assemblers.spades_default]
algorithm = "spades"
```

The next one is more organized and clustered for readability. I think this one is better.

```
[global_settings.filtering]
enabled = true                          # if false, proceed to downsample
min_length = 100                        # adjust for long reads   
quality = 15                            # adjust for long reads  

[global_settings.downsampling]
method = "none"                         # if method="none", proceed to assembly; "none"|"random"|"bbnorm"
target_num_reads = 0                    # adjust when method="random"; if target_num_reads=0, auto calculate target_num_reads using genome_size and target_depth
genome_size = 0                         # adjust when method="random" and target_num_reads=0; if genome_size=0, auto calculate genome size with MASH
target_depth = 150                      # adjust when method="random" and target_num_reads=0

[samples.sample1]
illumina = "short_reads_?.fastq.gz"     # glob path

[assemblers.spades_default]
algorithm = "spades"
```

For both of them, comments needs some updating but gives a good general idea of what users should input for each key and value.

standage:
I agree: the syntactic structure of the second example is a better reflection of the semantic structure of the config.

danejo3:
Cool. Internally, YEAT’s config is designed around a single config object that contains three classes.

The first class is `GlobalSettings`, the second is `Samples`, and the third is `Assemblers`. The config object contains only one `GlobalSettings` instance, but it can include multiple `Sample` and `Assembler` objects. In the proposed update, the GlobalSettings class is split into two components—either dictionaries ("filtering" and "downsampling") or nested objects. I’m opting for the object-based approach, so the GlobalSettings class will contain two required inner classes.

YEAT also has a useful feature that allows users to override `GlobalSettings` values directly within a `Sample` by redefining the same keys. I’m considering keeping this behavior rather than forcing a more structured nesting approach. For example, users could write:

```
[samples.sample1]
illumina = "data/short_reads_?.fastq.gz"   
downsample = "random"                        
target_num_reads = 0                        
genome_size = 4600000                           
target_depth = 150    
```

instead of:

```
[samples.sample1]
illumina = "data/short_reads_?.fastq.gz"    

[samples.sample1.downsampling]
downsample = "random"                        
target_num_reads = 0                        
genome_size = 4600000                           
target_depth = 150        
```                 

However, I’m still undecided. The latter is more explicit and structured, but it also makes the config file more verbose. Alternatively, we could support both styles: allow flat overrides for convenience while encouraging the nested structure for clarity and consistency.

danejo3:
Update: Karla prefers the latter approach, where users must be explicit. She pointed out that this helps preserve the conceptual separation between configuration domains (e.g., filtering vs. downsampling) and makes it clearer where each parameter belongs.

yeat/cli/cli.py
```diff
@@ -112,19 +110,23 @@ def grid_configuration(parser):
...
+[samples.sample1]
+illumina = "data/short_reads_?.fastq.gz"    # glob path
+
+[assemblers.spades_default]
+algorithm = "spades"'''
 
     def __call__(self, parser, namespace, values, option_string=None):
-        toml.dump(self.config_template, sys.stdout)
+        print(self.config_template)
         raise SystemExit()
```
danejo3, line 132: Here is the current `--init` output. I’ve updated some of the TOML keys to make them more explicit—for example, changing “downsampling” to “method” and “skip_filter” to “enabled.” The comments on the right side of each entry are included to help users understand what each parameter does. My goal is for the keys to be self-explanatory, with the comments serving as a reminder of when and how they’re applied. I’d appreciate any feedback or suggestions on how to improve this further.

```
[global_settings.filter]
enabled = true                              # if false, proceed to downsample
min_length = 100                            # adjust for long reads   
quality = 15                                # adjust for long reads  

[global_settings.downsample]
method = "none"                             # if method="none", proceed to assembly; "none"|"random"|"bbnorm"
target_num_reads = 0                        # adjust when downsample="random"; if target_num_reads=0, auto calculate input value using genome_size and target_depth
genome_size = 0                             # adjust when downsample="random" and target_num_reads=0; if genome_size=0, auto calculate input value using MASH
target_depth = 150                          # adjust when downsample="random" and target_num_reads=0

[samples.sample1]
illumina = "data/short_reads_?.fastq.gz"    # glob path

[assemblers.spades_default]
algorithm = "spades"
```

danejo3: Notice that the `global_settings` entry is now split into two separate sections. The goal is to make it clearer how each key applies to either the filter or downsample sections of the TOML.

Previously, users could override `global_settings` values applied to a specific sample by specifying the key and desired value directly within a sample's block.

For example, to disable filtering for sample1 and apply `bbnorm`, you could write:

```
[samples.sample1]
illumina = "data/short_reads_?.fastq.gz"
enabled = false
method = "bbnorm"
```

However, since the filter and downsample settings are now separated for clarity, users are required to define these options within their respective subsections rather than placing them at the top level of each sample.

For example:

```
[samples.sample1]
illumina = "data/short_reads_?.fastq.gz"

[samples.sample1.filter]
enabled = false

[samples.sample1.downsample]
method = "bbnorm"
```

This structure keeps configuration more explicit and organized, while still allowing per-sample customization. The biggest downside to doing this is the increased size of the TOML file, but I think the trade-off is worth it.

yeat/cli/just_yeat_it.py
```diff
@@ -47,15 +46,11 @@ def positional_args(parser):
...
         "read",
         nargs="+",
-        type=absolute_path,
+        type=lambda p: str(Path(p).resolve()),
         help="reads in FASTQ format; provide 2 for paired; provide 1 for single",
     )
 
 
-def absolute_path(path_str):
-    return str(Path(path_str).resolve())
```
danejo3, line 56: Reduce code space and use the lambda feature.

yeat/cli/just_yeat_it.py
```diff
@@ -66,66 +61,79 @@ def options(parser):
...
+        help="enable read filtering (default: enabled)",
     )
-    illumina.add_argument(
+    group.add_argument(
+        "--no-filter",
+        dest="filter",
+        action="store_false",
+        help="disable read filtering (ignores -l, -q)",
+    )
+    parser.set_defaults(filter_enabled=True)
```
danejo3, line 78: For `just-yeat-it`, all available options for configuring YEAT’s workflow are exposed. To make things more explicit and straightforward, I introduced a mutually exclusive group for enabling or disabling filtering. Previously, this was controlled by the `--skip_filter` flag, which would be set to `True` when filtering was skipped. To better align with the TOML terminology (i.e., using “enabled”), I opted for this updated approach instead.

yeat/cli/cli.py
danejo3: New to this TOML file is the ability to update the q-score for filtering for both short and long reads. Notice that by default, the value is 15. Users will need to manually adjust this value for long reads.

```
[global_settings.filter]
enabled = true                              # if false, proceed to downsample
min_length = 100                            # adjust for long reads   
quality = 15                                # adjust for long reads 
```

Users can "update" the quality score for filtering by overwriting or specifying the quality score in the sample section.

For example:

```
[samples.sample1]
illumina = "data/short_reads_?.fastq.gz"

[samples.sample2]
ont_duplex = "data/ont_reads.fastq.gz"

[samples.sample2.filter]
quality = 5
```

yeat/config/assemblers/assembler.py
```diff
@@ -9,12 +9,12 @@
...
 from pydantic import BaseModel, field_validator
-from typing import Optional, Dict
+from typing import Dict
 
 
 class Assembler(BaseModel):
     label: str
-    arguments: Optional[str]
+    arguments: str | None = None
     samples: Dict[str, Sample]
```
danejo3, line 18: Notice that I’ve updated the syntax from `Optional[...]` to using the `|` operator for the Pydantic class fields. The previous approach is now deprecated, and `|` is the preferred way to indicate that a parameter is optional when instantiating the class.

yeat/config/global_settings.py
```diff
@@ -7,19 +7,60 @@
...
 class GlobalSettings(BaseModel):
     model_config = ConfigDict(extra="forbid")
-    target_coverage_depth: Optional[int] = 150
-    target_num_reads: Optional[int] = -1  # -1 disable, 0 auto
-    genome_size: Optional[int] = 0  # 0 auto
-    min_length: Optional[int] = 100
-    quality: Optional[int] = 10
-    skip_filter: Optional[bool] = True
+    filter: FilterSettings
+    downsample: DownsampleSettings
```
danejo3, line 60: Consolidated the filter and downsample parameters/fields into their own individual pydantic classes.

yeat/config/global_settings.py
```diff
@@ -7,19 +7,60 @@
...
+
+class FilterSettings(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+    enabled: bool = True
+    min_length: int = 100
+    quality: int = 15
+
+    @classmethod
+    def parse_data(cls, data):
+        return cls(**data)
```
danejo3, line 22: Notice that we unpack the dictionary when creating the class. The `model_config = ConfigDict(extra="forbid")` makes checking extra dictionary keys or spelling errors easy to validate!

yeat/config/global_settings.py
```diff
@@ -7,19 +7,60 @@
...
+
+    def update(self, data):
+        extra_keys = set(data.keys()) - set(type(self).model_fields.keys())
+        if extra_keys:
+            raise FilterSettingsError(f"Extra field(s): {', '.join(extra_keys)}")
+        return self.model_copy(update=data)
```
danejo3, line 28: The purpose of this function is to update the class's field data. Each `Sample` object has a copy of the `GlobalSettings` object. When a user wants to overwrite a global setting value for a specific sample, this is where we update the `GlobalSettings` copy object within the `Sample` class. Unfortunately, we cannot use the `model_config = ConfigDict(extra="forbid")` but must manually check for extra keys or spelling errors.

yeat/config/sample.py
```diff
@@ -20,7 +21,9 @@
 
 class Sample(BaseModel):
     label: str
-    data: Dict[str, Union[list[Path], int, bool]]
+    data: Dict[str, list[Path]]
+    filter_settings: FilterSettings = FilterSettings()
+    downsample_settings: DownsampleSettings = DownsampleSettings()
```
danejo3, line 26: Notice that each `Sample` object has a `FilterSettings` and `DownsampleSettings` object. These objects are copies that represent the input keys and values from the `global_settings` section of the TOML file.

yeat/config/sample.py
```diff
@@ -31,42 +34,44 @@ def has_one_read_type(cls, data):
...
+        filter_copy = deepcopy(global_settings.filter).update(data.get("filter", {}))
+        downsample_copy = deepcopy(global_settings.downsample).update(data.get("downsample", {}))
+        data.pop("filter", None)
+        data.pop("downsample", None)
+        return cls(
+            label=label,
+            data=data,
+            filter_settings=filter_copy,
+            downsample_settings=downsample_copy,
+        )
```
danejo3, line 68: When creating the `Sample` object, we copy the `GlobalSettings` object `FilterSettings` and `DownsampleSettings`. After we copy the objects, we then update the fields if provided by the user. Afterwards, we then create the `Sample` object.

yeat/config/sample.py
```diff
@@ -31,42 +34,44 @@ def has_one_read_type(cls, data):
...
+
+    @model_validator(mode="after")
+    def has_valid_downsample_application(self):
+        if "illumina" not in self.data and self.downsample_settings.method == "bbnorm":
+            raise SampleConfigurationError("BBNorm can only be applied to Illumina reads")
+        return self
```
danejo3, line 74: Instead of doing a `post_init()` function, we can validate the model's fields after the object is created by doing it this way pydnatically. Here, we ensure that long-read samples with the `bbnorm` downsample instruction is not allowed as bbnorm is made for short-reads only.

yeat/config/sample.py
```diff
@@ -85,35 +90,39 @@ def has_long_reads(self):
...
+        for read_type in BEST_LR_ORDER:
+            if read_type in self.data:
+                return read_type
+        return None
 
     @property
-    def genome_size(self):
-        return self.data.get("genome_size", 0)
+    def enabled(self):
+        return self.filter_settings.enabled
```
danejo3, line 101: In this class, we have a bunch of `@property` that gets all of filtering and downsampling parameter values individually. No need for `.get()`. We can access this data directly through their class objects.

yeat/tests/data/configs/hybrid.toml
```diff
@@ -1,3 +1,6 @@
+[global_settings.filter]
+enabled = false
+
 [samples.hybrid_Shigella_sonnei_53G]
 illumina = "yeat/tests/data/short_reads_?.fastq.gz"
 pacbio_hifi = "yeat/tests/data/long_reads_high_depth.fastq.gz"
```
danejo3, line 6: This surfaced an interesting issue I hadn’t considered before. The `filter` and `downsample` settings are currently applied to all read types within a sample. This can be problematic for samples that contain both short and long reads, since the same filtering or downsampling parameters may not be appropriate for both. For samples that contain only a single read type (e.g., all long reads), this approach works fine.

At the moment, users working with hybrid samples who need different filtering or downsampling behavior will have to disable these steps and handle them manually.

One potential improvement would be to allow users to specify separate filtering and downsampling instructions for short- and long-read data within the same sample.

Maybe something like this?

```
[samples.hybrid_Shigella_sonnei_53G]
illumina = "yeat/tests/data/short_reads_?.fastq.gz"
pacbio_hifi = "yeat/tests/data/long_reads_high_depth.fastq.gz"

[samples.hybrid_Shigella_sonnei_53G.illumina.filter]
enabled = true

[samples.hybrid_Shigella_sonnei_53G.pacbio_hifi.filter]
enabled = false
```

yeat/workflow/qc/aux.py
```diff
@@ -7,12 +7,26 @@
...
-        copy(input, output)
+def copy_input(src, dst, do_copy):
+    if not is_gzip(src):
+        compress(src, dst)
         return
-    Path(output).symlink_to(input)
+    if do_copy:
+        copy(src, dst)
+    else:
+        Path(dst).symlink_to(src)
```
danejo3, line 22: I received a number of requests to allow YEAT to accept uncompressed FASTQ files. Originally, YEAT required users to compress their reads before running the workflow.

To address this, I’ve added support for both compressed and uncompressed inputs. If an input file is not compressed, YEAT will compress it into the sample’s working output directory rather than creating a symlink. This ensures that all files used by YEAT are consistently compressed and follow the expected filename conventions required by the Snakemake workflow.

yeat/workflow/qc/Long.smk
```diff
@@ -51,29 +51,79 @@ rule chopper:
...
+        genome_size=lambda wc: config["asm_cfg"].get_sample_genome_size(wc.sample),
+    log:
+        "analysis/{sample}/qc/{platform}/mash/mash.log",
+    run:
+        if params.genome_size:
+            shell("touch {output.mash_sentinel}")
+            return
+        shell("mash sketch -m {params.min_copies} -r {input.read} -o {params.sketch} > {log} 2>&1")
+        shell("mash info -t {params.sketch} > {params.mash_report}")
+        shell("touch {output.mash_sentinel}")
```
danejo3, line 86: To estimate genome size from the sample reads, YEAT uses MASH. To determine how many reads to retain during downsampling, it relies on the estimated genome size, the average read length, and a target coverage depth (150× by default).

To obtain the read length, I now use `seqkit` for both short- and long-read data, calculating the average read length. Previously, this information was extracted from `fastp` reports, but that approach only applies to short reads.

This rule could likely benefit from some refactoring. Currently, it generates a sentinel file as its primary output. The rationale is that if the genome size has already been provided or calculated, there’s no need to rerun MASH, and the sentinel file allows the workflow to skip this step accordingly.

yeat/workflow/qc/Long.smk
```diff
@@ -51,29 +51,79 @@ rule chopper:
...
+    input:
+        read=rules.chopper.output.read,
+    output:
+        seqkit_report="analysis/{sample}/qc/{platform}/seqkit/report.tsv",
+    wildcard_constraints:
+        platform="ont_simplex|ont_duplex|ont_ultralong|pacbio_hifi",
+    shell:
+        """
+        seqkit stats {input} > {output.seqkit_report}
+        """
```
danejo3, line 99: The purpose of this rule is to get the average read length.

yeat/workflow/qc/downsample.py
```diff
@@ -7,45 +7,45 @@
...
+from pathlib import Path
 
 
-class Downsample(BaseModel):
+@dataclass
+class Downsample:
+    target_num_reads: int
     genome_size: int
+    target_depth: int
     average_read_length: int
```
danejo3, line 20: The purpose of this class is to calculate the estimated number of reads to downsample. If the user provides the target number, use that instead.

yeat/workflow/qc/Paired.smk
```diff
@@ -55,53 +55,82 @@ rule fastp:
...
         min_length=lambda wc: config["asm_cfg"].get_sample_min_length(wc.sample),
+        quality=lambda wc: config["asm_cfg"].get_sample_quality(wc.sample),
     run:
-        if params.skip_filter:
+        if not params.filter_enabled:
             Path(output.r1).symlink_to(params.symlink_r1)
             Path(output.r2).symlink_to(params.symlink_r2)
             return
-        cmd = "fastp -i {input.r1} -I {input.r2} -o {output.r1} -O {output.r2} -l {params.min_length} --detect_adapter_for_pe --html {params.html_report} --json {params.json_report} 2> {params.txt_report}"
+        cmd = "fastp -i {input.r1} -I {input.r2} -o {output.r1} -O {output.r2} -l {params.min_length} -q {params.quality} --detect_adapter_for_pe --html {params.html_report} --json {params.json_report} 2> {params.txt_report}"
```
danejo3, line 66: Added a new parameters where users can adjust the q-score when filtering for short reads. By default, fastp sets it to 15 (the same for YEAT).

yeat/workflow/qc/Long.smk
```diff
@@ -51,29 +51,79 @@ rule chopper:
...
+        "analysis/{sample}/qc/{platform}/downsample/bbnorm.log",
     run:
-        if params.target_num_reads == -1:
+        if params.downsample_method == "none":
             Path(output.read).symlink_to(params.symlink_read)
             return
-        shell("seqtk sample -s {params.seed} {input.read} {params.target_num_reads} | gzip > {output.read}")
+        downsample = Downsample.parse_data(params.target_num_reads, params.genome_size, params.target_depth, params.mash_report, output.seqkit_report)
+        num_reads = downsample.get_num_reads(paired=False)
+        shell("seqtk sample -s {params.seed} {input.read} {num_reads} | gzip > {params.outdir}/read.fastq.gz")
```
danejo3, line 129: Reminder: No `bbnorm` application for long reads. Users can provide the target number to downsample with seqtk or have YEAT auto calculate it.

danejo3 (commented):
There’s quite a bit going on in this PR. The major changes include:

1. Updated help messages to make them clearer and more explicit in CLI.
2. Reorganized and expanded the TOML structure to make it more intuitive and user-friendly.
3. Refactored the configuration classes, introducing a `AssemblyConfig` class with supporting classes like `FilterSettings` and `DownsampleSettings`.
4. Added and revised several TOML keys, along with updated default values.
5. Introduced support for uncompressed FASTQ files.
6. Expanded test coverage beyond smoke tests to improve reliability.
7. Added support for applying BBNorm to Illumina reads during downsampling.
8. Cleaned up deprecated code and improved overall class structure.

This PR is ready for review. I’m also happy to walk through the changes in person, since the changelog alone might not fully capture the intent behind them. @standage

danejo3:
I realized after running several tests that both the proposed TOML format and the current format are not well suited for handling hybrid samples. Specifically, when a single sample contains both short and long reads, applying the same filtering or downsampling parameters across all read types becomes problematic.

For example, consider the following configuration:
```
[global_settings.filter]
enabled = true                              # if false, proceed to downsample
min_length = 100                            # adjust for long reads   
quality = 15                                # adjust for long reads 

[samples.sample1]
illumina = "data/short_reads_?.fastq.gz"    # glob path
ont_simplex = "data/long_reads_high_depth.fastq.gz"

[samples.sample1.filter]
quality = 9
```

In this case, the global filter settings are clearly optimized for short reads. However, if I keep the quality threshold at 15, I would likely lose most—or all—of the long-read data during filtering. On the other hand, lowering the quality threshold to 9 allows me to retain the long reads, but forces me to accept lower-quality Illumina reads than I would typically want. This highlights a fundamental limitation: a single set of filtering parameters cannot appropriately handle multiple read types within the same sample.

Because of this, users are effectively faced with two choices:

1. **Manual preprocessing**: Users can independently filter their read sets outside of the pipeline, disable filtering and downsampling in the TOML configuration, and then provide the processed inputs.
2. **Extend the TOML schema**: We can modify the configuration format to support read-type-specific parameters within a sample.

At this point, I lean toward the first option. It preserves the simplicity and intuitiveness of the current TOML structure, which I think is one of its strongest advantages. Introducing more specificity into the configuration risks making it harder to understand, maintain, and document--which goes completely against this PR.

That said, if we were to pursue the second option, a possible approach would look like this:

```
[samples.sample1.reads.illumina]
path = "data/short_reads_?.fastq.gz"    # glob path
filter = { min_length = 50 }

[samples.sample1.reads.ont_simplex]
path = "data/long_reads_high_depth.fastq.gz"
downsample = { method = "random", target_depth = 100 }
```

In this revised structure, we introduce a `reads` key under each sample, followed by entries for each read type. Each read type can then define its own `filter` and `downsample` parameters using inline dictionaries. This makes the configuration more expressive and allows fine-grained control over multiple input read types.

However, this added flexibility comes at a cost. The schema becomes more verbose and introduces additional nesting, which undermines the original goal of keeping the TOML format as simple and approachable as possible.

yeat/tests/data/configs/hybrid.toml
danejo3: Look at the most recent comment: https://github.com/bioforensics/yeat/pull/109#issuecomment-4224706365.

The example shown above does not work because TOML does not allow you to redefine existing an key. In this case, "illumina" and "pacbio_hifi".

danejo3:
[example_configs.pptx](https://github.com/user-attachments/files/26679985/example_configs.pptx)

Attached is a PowerPoint that showcases the new TOML format. Note: the YEAT workflow has not changed—only the required TOML format.

danejo3:
I didn't include this slide but we can for hybrid samples. This slide showcases the need for users to manually filter and downsample and not use YEAT for it.

Screenshot 2026-04-13 at 11 10 02 AM

standage:
Hey @danejo3, I think we can maintain flexibility even for hybrid assembly without too much complication. Consider the following as a candidate for the `yeat --init` output.

```toml
[global_settings.filter]
enabled = true
fastp_args = "--length_required 100 --unqualified_percent_limit 25"
chopper_args = "--quality 15 --minlength 250"

[global_settings.downsample.short]
enabled = true
method = "random"          # random|bbnorm
target_depth = 150         # desired depth of coverage
target_num_reads = "auto"  # number or "auto" to calculate from target_depth and genome_size; ignored if mode = "bbnorm"
genome_size = "auto"       # number (if genome size known) or "auto" to estimate using Mash; ignored if mode = "bbnorm"

[global_settings.downsample.long]
enabled = true
target_depth = 150         # desired depth of coverage
target_num_reads = "auto"  # number or "auto" to calculate from target_depth and genome_size
genome_size = "auto"       # number (if genome size known) or "auto" to estimate using Mash

[samples.sample1]
illumina = "data/sample1_R?.fastq.gz"

[assemblers.assembly1]
algorithm = "spades"
```

Overrides like so.

```toml
[global_settings.filter]
enabled = true
fastp_args = "--length_required 100 --unqualified_percent_limit 25"
chopper_args = "--quality 15 --minlength 250"

[global_settings.downsample.short]
enabled = true
method = "random"          # random|bbnorm
target_depth = 150         # desired depth of coverage
target_num_reads = "auto"  # number or "auto" to calculate from target_depth and genome_size; ignored if mode = "bbnorm"
genome_size = "auto"       # number (if genome size known) or "auto" to estimate using Mash; ignored if mode = "bbnorm"

[global_settings.downsample.long]
enabled = true
target_depth = 150         # desired depth of coverage
target_num_reads = "auto"  # number or "auto" to calculate from target_depth and genome_size
genome_size = "auto"       # number (if genome size known) or "auto" to estimate using Mash

[samples.sample1]
illumina = "data/sample1_R?.fastq.gz"

[samples.sample2]
illumina = "data/sample2_R?.fastq.gz"
pacbio_hifi = "data/sample2_hifi.fastq.gz"

[samples.sample3]
illumina = "data/sample3_R?.fastq.gz"
pacbio_hifi = "data/sample3_hifi.fastq.gz"
filter.chopper_args = "--quality 12 --minlength 200"
downsample.long.enabled = false

[assemblers.assembly1]
algorithm = "spades"

[assemblers.assembly2]
algorithm = "unicycler"
```

I'd be happy to chat more here or in person.

danejo3:
> Hey @danejo3, I think we can maintain flexibility even for hybrid assembly without too much complication. Consider the following as a candidate for the `yeat --init` output.
> 
> ```toml
> [global_settings.filter]
> enabled = true
> fastp_args = "--length_required 100 --unqualified_percent_limit 25"
> chopper_args = "--quality 15 --minlength 250"
> 
> [global_settings.downsample.short]
> enabled = true
> method = "random"          # random|bbnorm
> target_depth = 150         # desired depth of coverage
> target_num_reads = "auto"  # number or "auto" to calculate from target_depth and genome_size; ignored if mode = "bbnorm"
> genome_size = "auto"       # number (if genome size known) or "auto" to estimate using Mash; ignored if mode = "bbnorm"
> 
> [global_settings.downsample.long]
> enabled = true
> target_depth = 150         # desired depth of coverage
> target_num_reads = "auto"  # number or "auto" to calculate from target_depth and genome_size
> genome_size = "auto"       # number (if genome size known) or "auto" to estimate using Mash
> 
> [samples.sample1]
> illumina = "data/sample1_R?.fastq.gz"
> 
> [assemblers.assembly1]
> algorithm = "spades"
> ```
> 
> Overrides like so.
> 
> ```toml
> [global_settings.filter]
> enabled = true
> fastp_args = "--length_required 100 --unqualified_percent_limit 25"
> chopper_args = "--quality 15 --minlength 250"
> 
> [global_settings.downsample.short]
> enabled = true
> method = "random"          # random|bbnorm
> target_depth = 150         # desired depth of coverage
> target_num_reads = "auto"  # number or "auto" to calculate from target_depth and genome_size; ignored if mode = "bbnorm"
> genome_size = "auto"       # number (if genome size known) or "auto" to estimate using Mash; ignored if mode = "bbnorm"
> 
> [global_settings.downsample.long]
> enabled = true
> target_depth = 150         # desired depth of coverage
> target_num_reads = "auto"  # number or "auto" to calculate from target_depth and genome_size
> genome_size = "auto"       # number (if genome size known) or "auto" to estimate using Mash
> 
> [samples.sample1]
> illumina = "data/sample1_R?.fastq.gz"
> 
> [samples.sample2]
> illumina = "data/sample2_R?.fastq.gz"
> pacbio_hifi = "data/sample2_hifi.fastq.gz"
> 
> [samples.sample3]
> illumina = "data/sample3_R?.fastq.gz"
> pacbio_hifi = "data/sample3_hifi.fastq.gz"
> filter.chopper_args = "--quality 12 --minlength 200"
> downsample.long.enabled = false
> 
> [assemblers.assembly1]
> algorithm = "spades"
> 
> [assemblers.assembly2]
> algorithm = "unicycler"
> ```
> 
> I'd be happy to chat more here or in person.

This looks great! I really like how it simplifies the TOML file even further. I also appreciate how you defined and adjusted the filter arguments. I think `downsample.[long|short]` is probably the best way to handle our hybrid situation as well. I’m following what you’re doing. Let me see what I can come up with! Thanks for thinkin about this!

danejo3:
A slight update to the proposed TOML file configuration.

```
[global_settings.filter.short]
enabled = true
fastp_args = "--length_required 100 --unqualified_percent_limit 25"

[global_settings.filter.long]
enabled = false
chopper_args = ""

[samples.sample3]
illumina = "data/short_reads_?.fastq.gz"
pacbio_hifi = "data/long_reads_high_depth.fastq.gz"
```

In this example, we can independently enable or disable filters for short and long reads. The approach proposed last week forces both to be either filtered or unfiltered. You can’t apply filtering to just one type of read. This limitation can be problematic. For instance, you might want to filter your Illumina reads, while avoiding filtering for PacBio HiFi data, since those reads are already highly accurate. I’ve found this capability necessary because I’ve encountered situations where I needed to filter Illumina reads while completely disabling filtering for Nanopore reads in order to retain all of those reads for assembly.

standage:
That makes sense!

yeat/cli/cli.py
```diff
@@ -112,19 +110,38 @@ def grid_configuration(parser):
...
+[samples.sample1]
+illumina = "data/short_reads_R?.fastq.gz"   # illumina|pacbio_hifi|ont_simplex|ont_duplex|ont_ultralong
+
+[samples.sample2]
+illumina = "data/short_reads_R?.fastq.gz"
+pacbio_hifi = "data/long_reads.fastq.gz"
+
+[assemblers.assembly1]
+algorithm = "spades"
+arguments = ""'''
```
danejo3: Here is the current example `--init` config file. I used multiline string here to save and display the comments/notes. I think this is the simpliest config file that encapsulates everything that a user can do. For example, editing the various global parameters, the different types of input reads, and hybrid mode.

yeat/cli/just_yeat_it.py
```diff
@@ -66,67 +61,66 @@ def options(parser):
...
-    illumina = parser.add_argument_group("filter configuration")
-    illumina.add_argument(
-        "--skip-filter",
+    filter = parser.add_argument_group("filter configuration")
+    filter.add_argument(
+        "--filter",
         action="store_true",
-        help="skip the filtering step; default is to filter with fastp or chopper",
+        help="enable read filtering (default: disabled)",
     )
```
danejo3, line 69: By default, filtering is disabled. When running `just-yeat-it`, users can enable filtering by using the `--filter` flag.

yeat/cli/just_yeat_it.py
```diff
@@ -135,41 +129,50 @@ def algorithm_configuration(parser):
...
+        type=str,
     )
     algorithm.add_argument(
         "--algorithm",
+        choices=["spades", "megahit", "unicycler"],
         default="spades",
-        help='substitute the default assembly algorithm with another algorithm; for example, "megahit" or "unicycler"; by default, "spades"',
-        metavar="STR",
+        help='assembly algorithm (default: "spades")',
     )
```
danejo3, line 141: Let the `argparse` library handle the required input here for `--algorithm` flag.

yeat/config/downsample_settings.py
```diff
@@ -0,0 +1,61 @@
...
+from pydantic import BaseModel, ConfigDict, PositiveInt
+from typing import Literal
+
+
+class DownsampleSettings(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+    enabled: bool = False
+    target_depth: PositiveInt = 150
+    target_num_reads: PositiveInt | Literal["auto"] = "auto"
+    genome_size: PositiveInt | Literal["auto"] = "auto"
```
danejo3, line 19: The purpose of this file is to define the downsampling classes. The `DownsampleGroup` class encapsulates both short- and long-read downsampling settings. The `DownsampleSettings` class serves as the base for all downsampling parameters. `ShortDownsampleSettings` and `LongDownsampleSettings` extend `DownsampleSettings`. The `ShortDownsampleSettings` class includes an additional field, `method`, which is not present in `LongDownsampleSettings`.

yeat/config/filter_settings.py
```diff
@@ -0,0 +1,56 @@
...
+# National Biodefense Analysis and Countermeasures Center (NBACC), a Federally Funded Research and
+# Development Center.
+# -------------------------------------------------------------------------------------------------
+
+from pydantic import BaseModel, ConfigDict
+
+
+class FilterSettings(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+    enabled: bool = False
```
danejo3, line 15: Similar to the downsampling classes, this file defines the filtering classes.

The `FilterGroup` class encapsulates both short- and long-read filter settings. The `FilterSettings` class serves as the base for all filtering parameters. `ShortFilterSettings` and `LongFilterSettings` extend `FilterSettings`. The `ShortFilterSettings` class includes an additional field, `fastp_args`, which is not present in `LongFilterSettings`, while `LongFilterSettings` includes `chopper_args`, which is not present in `ShortFilterSettings`.

yeat/config/global_settings.py
```diff
@@ -7,19 +7,26 @@
...
 class GlobalSettings(BaseModel):
     model_config = ConfigDict(extra="forbid")
-    target_coverage_depth: Optional[int] = 150
-    target_num_reads: Optional[int] = -1  # -1 disable, 0 auto
-    genome_size: Optional[int] = 0  # 0 auto
-    min_length: Optional[int] = 100
-    quality: Optional[int] = 10
-    skip_filter: Optional[bool] = True
+    filter: FilterGroup = FilterGroup()
+    downsample: DownsampleGroup = DownsampleGroup()
```
danejo3, line 18: As mentioned above, we have two “Group” classes that define filtering and downsampling for both short and long reads. Instead of listing all fields directly, we encapsulate them within dedicated classes.

yeat/workflow/qc/downsample.py
```diff
@@ -7,45 +7,51 @@
...
+
+    def __post_init__(self):
+        if self.target_num_reads:
+            return
+        avl = (
+            2 * self.average_read_length
+            if self.read_type == "paired"
+            else self.average_read_length
+        )
+        self.target_num_reads = int((self.genome_size * self.target_depth) / avl)
```
danejo3, line 31: After this class is created, `target_num_reads` is always calculated, whether the user provides an explicit value or YEAT computes it automatically.

yeat/cli/just_yeat_it.py
```diff
@@ -66,67 +61,66 @@ def options(parser):
...
 
 def downsample_configuration(parser):
-    illumina = parser.add_argument_group("downsample configuration")
-    illumina.add_argument(
+    downsample = parser.add_argument_group("downsample configuration")
+    downsample.add_argument(
+        "--downsample",
+        action="store_true",
+        help="enable downsampling (default: disabled)",
+    )
```
danejo3, line 85: By default, downsampling is disabled. When running 'just-yeat-it', users can enable downsampling by using the `--downsample` flag.

danejo3 (commented):
@standage This PR is ready for review. It is quite large, so I recommend reviewing the files directly to understand the overall structure and how the changes fit together, in addition to looking at the diff. Many of the updates in this MR improve code clarity and organization.

Key Updates
- Updated the CLI help documentation for improved clarity and conciseness.
- Introduced new flags in `just-yeat-it`, including `--filter`, `--fastp_args`, `--downsample`, and `--method`.
- By default, YEAT disables filtering and downsampling in this workflow. Users can enable these features either in the TOML configuration file or via the appropriate `just-yeat-it` flags.
- Updated the `--init` example.
- Migrated Pydantic class syntax to v3.
- Introduced an object-oriented approach for handling filtering and downsampling for short and long reads.
- Revamped the required TOML configuration file and added comments to guide users.
- Improved tests to be more targeted and less reliant on smoke testing.
- Increased test coverage to 73%. The remaining uncovered areas are primarily functions in the assembler classes, which would significantly expand the test suite and require substantial effort to cover.
- Added the ability to downsample reads either randomly using `seqtk` or via digital normalization using `bbnorm`.

danejo3:
Here's the updated TOML examples in powerpoint!

[example_configs.pptx](https://github.com/user-attachments/files/27130787/example_configs.pptx)

yeat/cli/cli.py
danejo3: See new and updated version here: https://github.com/bioforensics/yeat/pull/109#discussion_r3147474160

yeat/cli/just_yeat_it.py
danejo3: Outdated and not used anymore.

yeat/cli/just_yeat_it.py
```diff
@@ -135,41 +129,52 @@ def algorithm_configuration(parser):
...
-    data = get_config_data(args)
     workdir = Path(args.workdir)
     workdir.mkdir(parents=True, exist_ok=True)
-    outfile = open(workdir / "config.toml", "w")
-    toml.dump(data, outfile)
+    config = workdir / "config.toml"
+    with open(config, "w") as f:
+        data = get_config_data(args)
+        toml.dump(data, f)
+    setattr(args, "config", str(config.resolve()))
```
standage: If `args.config = str(config.resolve())` works here, it would be a more idiomatic way to perform this operation.

yeat/config/downsample_settings.py
```diff
@@ -0,0 +1,61 @@
...
+
+    @classmethod
+    def parse_data(cls, data):
+        return cls(**data)
+
+    def update(self, data):
+        extra_keys = set(data.keys()) - set(type(self).model_fields.keys())
+        if extra_keys:
+            raise DownsampleSettingsError(f"Extra field(s): {', '.join(extra_keys)}")
```
standage: Is this error handling necessary here? If the dictionary contains insufficient or unnecessary data, that will be caught when `model_copy` is executed, right? If so, the error handling is probably better situated where `.update()` is called.

yeat/config/filter_settings.py
```diff
@@ -0,0 +1,56 @@
...
+class FilterSettingsError(ValueError):
+    pass
+
+
+class ShortFilterSettings(FilterSettings):
+    fastp_args: str = ""
+
+
+class LongFilterSettings(FilterSettings):
+    chopper_args: str = ""
```
standage: I think it would be appropriate to include a reasonable default configuration for filter settings—and assembly settings for that matter. We can discuss what precisely these should be, but for fastp we should at least set a minimum length threshold. We may consider tweaking other options as well.

yeat/cli/just_yeat_it.py
```diff
@@ -135,41 +129,52 @@ def algorithm_configuration(parser):
...
-        metavar="STR",
+        help='assembly algorithm (default: "spades")',
     )
     algorithm.add_argument(
         "--arguments",
         default="",
-        help='add assembly algorithm flags; for example, "--meta" or "--isolate --careful" for SPAdes; by default, empty string',
+        help='additional assembly algorithm arguments; e.g., "--meta", "--isolate --careful" (default: none)',
         metavar="STR",
+        type=str,
```
standage: I think `--isolate` would be a reasonable default here.

yeat/workflow/qc/Long.smk
```diff
@@ -51,29 +52,77 @@ rule chopper:
...
+        genome_size=lambda wc: config["asm_cfg"].get_sample_genome_size(wc.sample, "long"),
+    log:
+        "analysis/{sample}/qc/{platform}/mash/mash.log",
+    run:
+        if isinstance(params.genome_size, int):
+            shell("touch {output.mash_sentinel}")
+            return
+        shell("mash sketch -m {params.min_copies} -r {input.read} -o {params.sketch} > {log} 2>&1")
+        shell("mash info -t {params.sketch} > {params.mash_report}")
+        shell("touch {output.mash_sentinel}")
```
standage: If you do...

```
output:
    mash_sentinel=touch("analysis/{sample}/qc/{platform}/mash/sentinel.done"),
```

...you could simplify the rule.

```python
if isinstance(params.genome_size, int):
    return
shell("mash sketch -m {params.min_copies} -r {input.read} -o {params.sketch} > {log} 2>&1")
shell("mash info -t {params.sketch} > {params.mash_report}")
```

yeat/workflow/qc/Paired.smk
```diff
@@ -55,53 +55,82 @@ rule fastp:
...
+    params:
+        min_copies=2,
+        sketch="analysis/{sample}/qc/illumina/mash/reference.msh",
         mash_report="analysis/{sample}/qc/illumina/mash/report.tsv",
+        genome_size=lambda wc: config["asm_cfg"].get_sample_genome_size(wc.sample, "short"),
+    log:
+        "analysis/{sample}/qc/illumina/mash/mash.log",
+    run:
+        if isinstance(params.genome_size, int):
+            shell("touch {output.mash_sentinel}")
```
standage: Sam `touch()` trick with this sentinel.

yeat/workflow/qc/Single.smk
```diff
@@ -48,47 +48,74 @@ rule fastp:
...
+        min_copies=2,
+        sketch="analysis/{sample}/qc/illumina/mash/reference.msh",
         mash_report="analysis/{sample}/qc/illumina/mash/report.tsv",
+        genome_size=lambda wc: config["asm_cfg"].get_sample_genome_size(wc.sample, "short"),
+    log:
+        "analysis/{sample}/qc/illumina/mash/mash.log",
+    run:
+        if isinstance(params.genome_size, int):
+            shell("touch {output.mash_sentinel}")
+            return
```
standage, line 75: Same `touch()` trick here.

standage (changes requested):
This is looking good! I've added a few questions, comments, and suggestions.

yeat/config/filter_settings.py
danejo3: I used your example fastp and chopper arguments as the defaults for YEAT. The values you provided seem reasonable and should work well in most cases, though users can adjust them manually if needed.

```
[global_settings.filter.short]
enabled = false
fastp_args = "--length_required 100 --unqualified_percent_limit 25"

[global_settings.filter.long]
enabled = false
chopper_args = "--quality 15 --minlength 250"
```

yeat/workflow/qc/Long.smk
danejo3: image

yeat/config/downsample_settings.py
danejo3: The reason for the check statement here is that `model_copy()` does not perform input validation, so I had to manually add a check. Input validation, as far as I understand, only occurs when the model is initially created. I originally looped through the input dictionary to check for invalid keys, but I later found a Pydantic-native way to update the object. However, this approach creates a shallow copy of the object (the one that you are seeing above!).

yeat/tests/data/configs/auto_one_sample.toml
```diff
@@ -3,3 +3,15 @@ illumina = "yeat/tests/data/short_reads_*.fastq.gz"
...
+[global_settings.filter.short]
+enabled = false
+fastp_args = "--length_required 100 --unqualified_percent_limit 25"
+
+[global_settings.downsample.short]
+enabled = false
+target_depth = 150
+target_num_reads = "auto"
+genome_size = "auto"
+method = "random"
```
danejo3, line 17: `yeat-auto` now includes default global setting values to the TOML file. Also, I've included `--isolate` in the spades' arguments section as well.

yeat/config/filter_settings.py
```diff
@@ -19,27 +19,24 @@ def parse_data(cls, data):
...
 
 class FilterGroup(BaseModel):
-    short_filter_settings: ShortFilterSettings = ShortFilterSettings()
-    long_filter_settings: LongFilterSettings = LongFilterSettings()
+    short_filter_settings: ShortFilterSettings = Field(
+        default_factory=ShortFilterSettings, serialization_alias="short"
+    )
+    long_filter_settings: LongFilterSettings = Field(
+        default_factory=LongFilterSettings, serialization_alias="long"
+    )
```
danejo3, line 41: I found a useful feature in Pydantic models. When calling `model_dump()`, the dictionary keys are based on the model’s field names. In my case, I deliberately used different field names instead of the original input TOML file keys to improve readability in the code, rather than using something like `short: ShortFilterSettings`. 

Pydantic allows you to define alias names for fields when serializing models. You can control whether these aliases are used during serialization by calling `model_dump(by_alias=True)`, which outputs the data using the alias names instead of the field names.

yeat/config/downsample_settings.py
```diff
@@ -23,14 +23,7 @@ def parse_data(cls, data):
...
     def update(self, data):
-        extra_keys = set(data.keys()) - set(type(self).model_fields.keys())
-        if extra_keys:
-            raise DownsampleSettingsError(f"Extra field(s): {', '.join(extra_keys)}")
-        return self.model_copy(update=data)
-
-
-class DownsampleSettingsError(ValueError):
-    pass
+        return type(self)(**{**self.model_dump(), **data})
```
danejo3: To ensure that the input data used to update the class does not contain any invalid keys, we can leverage Pydantic’s validation system by re-creating the object with the updated field values and returning the newly validated instance.

I've changed the update method from `model_copy()` to dereferencing the model and the data.

danejo3 (commented):
@standage This is ready for another review! Thanks for the tips and suggestions. They were great! I've added some comments to the questions that you had. Also, in the recent commit, I've updated yeat-auto to include the global filter and downsample settings to the TOML output string.

yeat/config/downsample_settings.py
danejo3: See the new change here: https://github.com/bioforensics/yeat/pull/109#discussion_r3161981944 .

danejo3:
[example_configs.pptx](https://github.com/user-attachments/files/27207103/example_configs.pptx)

yeat/config/downsample_settings.py
standage: I like that this approach takes advantage of pydantic's built-in validation and doesn't reinvent it. But we should consider unpacking the dense syntax to clarify what each bit of code is doing. Does something like this work?

```python
def update(self, data):
    updated_data = {**self.model_dump(), **data}
    self_type = type(self)
    return self_type(updated_data)
```

danejo3: I need to unpack the dictionary further; otherwise, I’d have to create an `__init__` that accepts a dictionary and would bypass self-validation.

```
def update(self, data):
        updated_data = {**self.model_dump(), **data}
        self_type = type(self)
        return self_type(**updated_data)
```

standage: You're right, that was an omission on my part!

yeat/config/filter_settings.py
danejo3: I spoke with the analysts, and they would like us to use the same default fastp parameters in another internal workflow. How do these look as defaults? The internal workflow does not use `--unqualified_percent_limit 25`, so we can simplify the defaults to: `--min-length 50 --detect_adapter_for_pe`. Long-read filtering will be replaced with our internal long-read filtering tool once Kyle is able to release it on GitHub (see #108).

```
[global_settings.filter.short]
enabled = false
fastp_args = "--min-length 50 --unqualified_percent_limit 25 --detect_adapter_for_pe"

[global_settings.filter.long]
enabled = false
chopper_args = "--quality 15 --minlength 250"
```

danejo3: Final version: we’ll stick with the same default parameters as the internal workflow. We won’t include filtering parameters for long reads, since the internal workflow assumes high read quality and does not apply filtering. Users will need to adjust the settings manually based on their needs.

```
[global_settings.filter.short]
enabled = false
fastp_args = "--min-length 50 --detect_adapter_for_pe"

[global_settings.filter.long]
enabled = false
chopper_args = ""
```

Note for `yeat --init`.

By default, the example output sets SPAdes’ extra arguments to `--isolate`. However, in the SPAdesAssembler class, the arguments field is an empty string. As a result, users will need to manually specify `--isolate` mode each time.

```
[assemblers.assembly1]
algorithm = "spades"
arguments = "--isolate"
```

All other assembly classes do not define an arguments field with default values. We could add `--isolate` mode as the default.

standage: I was just looking at this. Sounds good!

standage (approved):

PR #110, opened 2026-04-30
v0.9 Patch Release

danejo3:
Fix test and updated various files for another official release.

### DIFF
--- a/yeat/tests/test_just_yeat_it.py
+++ b/yeat/tests/test_just_yeat_it.py
@@ -54,10 +54,7 @@ def test_invalid_input_algorithm(capsys, tmp_path):
         data_file("short_reads_1.fastq.gz"),
         data_file("short_reads_2.fastq.gz"),
     ]
-    with pytest.raises(SystemExit, match="2"):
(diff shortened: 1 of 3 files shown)


yeat/tests/test_just_yeat_it.py
```diff
@@ -54,10 +54,7 @@ def test_invalid_input_algorithm(capsys, tmp_path):
         data_file("short_reads_1.fastq.gz"),
         data_file("short_reads_2.fastq.gz"),
     ]
-    with pytest.raises(SystemExit, match="2"):
+    with pytest.raises(SystemExit):
         run_yeat(arglist)
-    captured = capsys.readouterr()
-    message = (
-        "argument --algorithm: invalid choice: 'DNE' (choose from spades, megahit, unicycler)"
-    )
```
danejo3, line 62: Adjusted string comparison to account for OS-specific differences in argparse error messages. On macOS, choices appear without quotes (spades, megahit, unicycler), whereas on Linux they are quoted when using --algorithm. To make it easy, I spliced off the choices and just went with the most important substring to match.

PR #111, opened 2026-05-04
0.9.2 patch fix

danejo3:


yeat/cli/cli.py
```diff
@@ -112,7 +112,7 @@ def grid_configuration(parser):
 class InitAction(Action):
     config_template = '''[global_settings.filter.short]
 enabled = false
-fastp_args = "--min-length 50 --detect_adapter_for_pe"
+fastp_args = "--length_required 50 --detect_adapter_for_pe"
```
danejo3, line 115: The correct flag that should be used for `fastp` is `--length_required` or `-l`.

yeat/cli/cli.py
```diff
@@ -121,7 +121,7 @@ class InitAction(Action):
 [global_settings.downsample.short]
 enabled = false
 method = "random"                           # random|bbnorm
-target_depth = 150                          # desired depth of coverage
+target_depth = 150                          # desired depth of coverage; ignored if mode = "bbnorm"
```
danejo3: Analysts were unsure whether `target_depth` applied to `seqtk` in random mode, `bbnorm`, or both.

Although this wasn’t initially clear, `bbnorm` can adjust based on a “target depth.” By default, the target depth is set to 10×. Analysts have not changed this parameter and do not plan to adjust it for `bbnorm`.

While we could allow users to modify the target depth for `bbnorm`, doing so may introduce unnecessary complexity. We could add separate parameters such as `random_target_depth` for random downsampling and `bbnorm_target_depth` but... 💁. For now, we will not allow adjustment of `target_depth` for `bbnorm`.

danejo3 (commented):
More feedback and bug fixes.

yeat/cli/cli.py
standage: I’m not sure what you mean when you say it wasn’t initially clear that `bbnorm` has a target depth. If we’re talking about what I think we’re talking about, it’s a core part of the algorithm.

I’m surprised to hear that analysts have not overridden the 10x default. I’d like us to confirm this before proceeding. In any case, I’m not yet convinced that a single variable isn’t appropriate for both downsampling approaches.

danejo3: > I’m surprised to hear that analysts have not overridden the 10x default.

Correction—my fault, I missed an extra zero. It’s 100x by default.

> I’m not sure what you mean when you say it wasn’t initially clear that *bbnorm* has a target depth. If we’re talking about the same thing, that’s a core part of the algorithm.

Sorry for the confusion—I realized I confused myself. Please ignore what I said earlier...

`bbnorm` has a `target` flag, which is set to 100x by default. In their help message, this is described as “(tgt) Target normalization depth.” At the moment, users cannot modify the `target` value for `bbnorm`.

I need to update the Snakemake rule so that `bbnorm` can accept the user-provided `target_depth` input.

danejo3: I’ve spoken with the analysts. YEAT’s default `target_depth` is set to 150x, and we will keep it as-is. If users choose to run `bbnorm`, they can set the target depth to 100x to match `bbnorm`’s default value. 

I also asked whether they have been adjusting the `target` when running `bbnorm`, and they said this is infrequent. 

That said, we could reduce the default from 150x to 100x if needed—the change would be straightforward to implement.

danejo3:
`bbnorm` command updated with `target` for both single- and paired-end reads.

~~(Testing. Github commenting is not working?)~~

I can't comment my code for some reason. Github is having internal errors on my end.

yeat/cli/cli.py
standage: Ok, thanks for clarifying. If 100x is bbnorm's default, that's much more consistent with the behavior that I've discussed with colleagues recently.

standage:
Yes, Github is having a moment it would seem.

CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.9.2] 2026-05-04
+
+### Added
+- Ability to adjust target depth when using BBNorm (#111)
+
+### Changed
+- Incorrect fastp minimum read length argument (#111)
```
standage: I'd suggest both of these could go under the heading `### Fixed` instead.

```
### Fixed
- Applied `target_depth` downsampling setting to BBnorm (#111)
- Fastp minimum read length argument (#111)
```

And it's a new day 😃

danejo3:
@standage This is ready for review!

standage (approved):

